PR #684, opened 2015-01-18
Clarifying what happens in the "idle" stream state

### LINKED ISSUE #667: section 5.1 PRIORITY frame and idle state clarification
[section 5.1](https://tools.ietf.org/html/draft-ietf-httpbis-http2-16#section-5.1) has the following text:
>idle:
      All streams start in the "idle" state.  In this state, no frames
      have been exchanged.

It seems like PRIORITY frames can be sent/received in the IDLE state without causing a state transition as illustrated in [figure 2](http://http2.github.io/http2-spec/#StreamStatesFigure).  If I am interpreting this correctly then the `no frames have been exchanged` statement is not necessarily correct if PRIORITY frames have been exchanged.

martinthomson:
This removes the misleading text as observed in #667.  It also clarifies what was potentially misleading text about PUSH_PROMISE from the section.

Closes #667

### DIFF
--- a/draft-ietf-httpbis-http2.xml
+++ b/draft-ietf-httpbis-http2.xml
@@ -847,20 +847,20 @@ HTTP2-Settings    = token68
                     cause a stream to immediately become "half closed".
                   </t>
                   <t>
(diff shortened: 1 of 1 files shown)


martinthomson:
Manually merged

PR #685, opened 2015-01-18
It is a compression error to exceed the defined table size

### LINKED ISSUE #676: Mandate error for violation of maximum compression table size
I think that this is editorial, but we need to be super careful here.

See http://www.w3.org/mid/D0D94EF7.3E28E%25Robby.Simpson@GE.com

martinthomson:
Closes #676


hruellan:
Replaced by #705 for compatibility with other changes.

PR #687, opened 2015-01-18
Reset of promised stream for unsafe method clarifications

### LINKED ISSUE #679: New safe methods
> If a client receives a PUSH_PROMISE that does not include a complete 
and valid set of header fields, or the :method header field identifies a 
method that is not safe, it MUST respond with a stream error (Section 
5.4.2) of type PROTOCOL_ERROR.

New safe methods can be defined. Does this mean that the client needs to 
respond with a stream error for methods *not known to be safe*?

See http://www.w3.org/mid/54B53822.2020104@gmx.de

martinthomson:
Closes #679

### DIFF
--- a/draft-ietf-httpbis-http2.xml
+++ b/draft-ietf-httpbis-http2.xml
@@ -3333,9 +3333,11 @@ HTTP2-Settings    = token68
           Promised requests MUST be cacheable (see <xref target="RFC7231" x:fmt=","
           x:rel="#cacheable.methods"/>), MUST be safe (see <xref target="RFC7231" x:fmt=","
           x:rel="#safe.methods"/>) and MUST NOT include a request body. Clients that receive a
-          promised request that is not cacheable, unsafe or that includes a request body MUST
-          reset the stream with a <xref target="StreamErrorHandler">stream error</xref> of type
-          <x:ref>PROTOCOL_ERROR</x:ref>.
+          promised request that is not cacheable, unsafe or that indicates the presence of a request
+          body MUST reset the promised stream with a <xref target="StreamErrorHandler">stream
+          error</xref> of type <x:ref>PROTOCOL_ERROR</x:ref>.  Note this could result in the
+          promised stream being reset if the client does not recognize the a newly defined method as
+          being safe.
         </t>
         <t>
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2.xml
```diff
@@ -3333,9 +3333,11 @@ HTTP2-Settings    = token68
           Promised requests MUST be cacheable (see <xref target="RFC7231" x:fmt=","
           x:rel="#cacheable.methods"/>), MUST be safe (see <xref target="RFC7231" x:fmt=","
           x:rel="#safe.methods"/>) and MUST NOT include a request body. Clients that receive a
-          promised request that is not cacheable, unsafe or that includes a request body MUST
-          reset the stream with a <xref target="StreamErrorHandler">stream error</xref> of type
-          <x:ref>PROTOCOL_ERROR</x:ref>.
+          promised request that is not cacheable, unsafe or that indicates the presence of a request
+          body MUST reset the promised stream with a <xref target="StreamErrorHandler">stream
+          error</xref> of type <x:ref>PROTOCOL_ERROR</x:ref>.  Note this could result in the
+          promised stream being reset if the client does not recognize the a newly defined method as
```
mnot: English is awkward here - "the a"

draft-ietf-httpbis-http2.xml
```diff
@@ -3333,9 +3333,11 @@ HTTP2-Settings    = token68
           Promised requests MUST be cacheable (see <xref target="RFC7231" x:fmt=","
           x:rel="#cacheable.methods"/>), MUST be safe (see <xref target="RFC7231" x:fmt=","
           x:rel="#safe.methods"/>) and MUST NOT include a request body. Clients that receive a
-          promised request that is not cacheable, unsafe or that includes a request body MUST
-          reset the stream with a <xref target="StreamErrorHandler">stream error</xref> of type
-          <x:ref>PROTOCOL_ERROR</x:ref>.
+          promised request that is not cacheable, unsafe or that indicates the presence of a request
```
phluid61: "...that is not cachable, unsafe..." implies "is not unsafe." It's obvious in the context of the prose above, but "...is not cachable, is unsafe..." is less ambiguous.

PR #694, opened 2015-01-20
More additions for GenART review

hruellan:


draft-ietf-httpbis-header-compression.xml
```diff
@@ -567,6 +566,13 @@
                     entire table causes the table to be emptied of all existing
                     entries, and results in an empty table.
                 </t>
+                <t>
+                    A new entry can reference the name of an entry in the
+                    dynamic table that will be evicted when adding this new
+                    entry into the dynamic table. Implementations have to take
+                    care not deleting the referenced name from memory while
+                    evicting the referenced entry from the dynamic table.
```
martinthomson: Grammar is a little odd here.  Perhaps: 

Implementations are cautioned to avoid deleting the referenced name if the entry is evicted from the dynamic table prior to inserting the new entry.

PR #696, opened 2015-01-20
COMPRESSION_ERROR clarification

RobbySimpson:
Proposal to resolve issue #695

PR #699, opened 2015-01-21
Adding a failure note for decompression failure

martinthomson:
Proposal to replace #696.  I think that this is a little clearer this way.

Also, it's strange that this was not added earlier, but then sometimes the most obvious things get missed.

cc @RobbySimpson

### DIFF
--- a/draft-ietf-httpbis-http2.xml
+++ b/draft-ietf-httpbis-http2.xml
@@ -718,7 +718,7 @@ HTTP2-Settings    = token68
           Header compression is stateful.  One compression context and one decompression context is
           used for the entire connection.  Failure to decompress a header
           block MUST be treated as a <xref target="ConnectionErrorHandler">connection error</xref>
-          of type <x:ref>COMRPESSION_ERROR</x:ref>.
+          of type <x:ref>COMPRESSION_ERROR</x:ref>.
         </t>
         <t>
           Each header block is processed as a discrete unit.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2.xml
```diff
@@ -716,7 +716,12 @@ HTTP2-Settings    = token68
         </t>
         <t>
           Header compression is stateful.  One compression context and one decompression context is
-          used for the entire connection.  Each header block is processed as a discrete unit.
+          used for the entire connection.  Failure to decompress a header
+          block MUST be treated as a <xref target="ConnectionErrorHandler">connection error</xref>
+          of type <x:ref>COMRPESSION_ERROR</x:ref>.
```
ijl: @martinthomson, I see there's a typo on `COMRPESSION_ERROR`.

RobbySimpson:
So I'm not seeing how the new sentence is any different than this one two paragraphs below:
A receiver MUST terminate the connection with a connection error of type COMPRESSION_ERROR if it does not decompress a header block.

My chief concern (merely editorial) is that HPACK consistently uses the phrase "decoding error" whereas here we are using the phrase "decompress."  I think it good to make clear that any and all decoding errors MUST result in connection errors of type COMPRESSION_ERROR.

Picky, I know.  ;-)

hruellan:
I think that Martin's proposal tries to specifically address your concern.

The sentence two paragraphs below address the case of an endpoint deciding for whatever reason not to decompress a header block: this also has to result in a COMPRESSION_ERROR.

So I think that both are useful.

RobbySimpson:
Ahhh - thanks @hruellan!  Makes perfect sense and I agree.

PR #700, opened 2015-01-23
Reset state

gregw:
The complex conditional behaviour of the closed state indicates that it is really two states: reset and closed.  The reset state is entered when a reset is sent and then transitions to the closed state only when the peer acknowledges the stream is closed with either an END_STREAM flag or RST_STREAM frame.

By making RST_STREAM an acknowledged frame there is no need for indeterminate behaviour with timeouts to handle frames sent while a RST_STREAM is in transit.

### DIFF
--- a/draft-ietf-httpbis-http2.xml
+++ b/draft-ietf-httpbis-http2.xml
@@ -777,7 +777,7 @@ HTTP2-Settings    = token68
                     ,--------|  idle  |--------.
                    v         +--------+         v
            +----------+          |           +----------+
-     ,-----| reserved |          | semd H/   | reserved |-----.
+     ,-----| reserved |          | send H/   | reserved |-----.
      |     | (local)  |          | recv H    | (remote) |     |
      |     +----------+          |           +----------+     |
      |         |                 v                  |         |
--- a/draft-ietf-httpbis-http2.xml
+++ b/draft-ietf-httpbis-http2.xml
@@ -3675,40 +3675,30 @@ HTTP2-Settings    = token68
           </t>
         </section>
 
-        <section title="TLS 1.2 Cipher Suites">
+        <section title="TLS Cipher Suites">
           <t>
-            The set of TLS 1.2 cipher suites that are permitted in HTTP/2 is restricted.
-	  </t>
-	  <t>HTTP/2 MUST only be used with ciphers that exchange keys with either:
-            <xref target="TLS12">ephemeral Diffie-Hellman (DHE)</xref> with a minimum key 
-	    exchange size of 2048;  
-	    or the <xref target="RFC4492">elliptic curve variant (ECDHE)</xref> with a 
-	    security level of at least 128 bits. 
-            Clients MUST accept DHE key exchange sizes of up to 4096 bits.  
-	  </t>
-	  <t>
-            HTTP/2 MUST NOT be used with 
-            <xref target="RFC5246" >Null or Standard Stream Ciphers (Section 6.2.3.1)</xref> 
-            nor <xref target="RFC5246">CBC Block Ciphers (Section 6.2.3.3)</xref>.
+            The set of TLS cipher suites that are permitted in HTTP/2 is restricted.  HTTP/2 MUST
+            only be used with cipher suites that have ephemeral key exchange, such as the <xref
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2.xml
```diff
@@ -773,34 +773,34 @@ HTTP2-Settings    = token68
...
-     |         |             +--------+             |         |
-     |         |     recv ES |        | send ES     |         |
-     |  send H |     ,-------|  open  |-------.     | recv H  |
-     |         |    /        |        |        \    |         |
-     |         v   v         +--------+         v   v         |
-     |     +----------+          |           +----------+     |
-     |     |   half   |          |           |   half   |     |
-     |     |  closed  |          | send R/   |  closed  |     |
-     |     | (remote) |          | recv R    | (local)  |     |
+     ,-----| reserved |          | semd H/   | reserved |-----.
```
Scottmitch: Semd -> send

draft-ietf-httpbis-http2.xml
```diff
@@ -995,64 +995,64 @@ HTTP2-Settings    = token68
                 target="FlowControl">stream level flow control limits</xref>.
               </t>
               <t>
-                A stream can transition from this state to "closed" by sending a frame that contains
-                an END_STREAM flag, or when either peer sends a <x:ref>RST_STREAM</x:ref> frame.
+                A stream transitions from this state to "closed" when it receives  
+		an END_STREAM flag or a <x:ref>RST_STREAM</x:ref> frame. A stream transitions from this
```
Scottmitch: Indentation is a bit off

draft-ietf-httpbis-http2.xml
```diff
@@ -995,64 +995,64 @@ HTTP2-Settings    = token68
...
-                If this state is reached as a result of sending a <x:ref>RST_STREAM</x:ref> frame,
-                the peer that receives the <x:ref>RST_STREAM</x:ref> might have already sent - or
-                enqueued for sending - frames on the stream that cannot be withdrawn.  An endpoint
-                MUST ignore frames that it receives on closed streams after it has sent a
-                <x:ref>RST_STREAM</x:ref> frame.  An endpoint MAY choose to limit the period over
-                which it ignores frames and treat frames that arrive after this time as being in
-                error.
+                Flow controlled frames (i.e., <x:ref>DATA</x:ref>) received in this state
+                are counted toward the connection flow control window, even though their content
+		is ignored as the sender will consider the frames to count
```
Scottmitch: Indentation

draft-ietf-httpbis-http2.xml
```diff
@@ -995,64 +995,64 @@ HTTP2-Settings    = token68
...
+              </t>
+              <t>
+                An endpoint MUST NOT send frames other than <x:ref>PRIORITY</x:ref> on a closed
+                stream.  An endpoint that receives any frame other than <x:ref>PRIORITY</x:ref>
+                or <x:ref>RST_STREAM</x:ref> in this state MUST treat that as a 
+		<xref target="StreamErrorHandler">stream error</xref> of type
+                <x:ref>STREAM_CLOSED</x:ref>.  
+              </t>
+              <t>
+                <x:ref>PRIORITY</x:ref> frames can be sent on closed streams to prioritize streams
```
Scottmitch: Should `closed` be in quotes for consistency when referring to states?

gregw: Probably, but I noticed that it is other states are not consistenty quoted, so was following the practise of what was before me.... but will change to consistency as I think that is best.

gregw: Looks like the convention is to use "closed" when talking about the state, but just closed when talking about the stream.   Document is full of closed streams so I wont change those.

thanks

gregw:
I just noticed that my branch picked up a couple of reverted commits from previous PRs.   So IF this is accepted, I'll remake it on a fresh branch without the noise commits.  But I'll leave it as is for now to see if it has a hope first.

gregw:
I couldn't stand the mess, so I have recreated without noise reverts and with Scotts feedback in https://github.com/http2/http2-spec/pull/701

PR #701, opened 2015-01-23
Added Reset State

gregw:
The complex conditional behaviour of the closed state indicates that it is really two states: reset and closed. The reset state is entered when a reset is sent and then transitions to the closed state only when the peer acknowledges the stream is closed with either an END_STREAM flag or RST_STREAM frame.

By making RST_STREAM an acknowledged frame there is no need for indeterminate behaviour with timeouts to handle frames sent while a RST_STREAM is in transit.

This is a redo of PR#700 on a cleaner branch without reverted commits.


draft-ietf-httpbis-http2.xml
```diff
@@ -773,34 +773,34 @@ HTTP2-Settings    = token68
...
+     |         |                 v                  |         |
+     |         |     recv ES +--------+ send ES     |         |
+     |  send H |     ,-------|  open  |-------.     | recv H  |
+     |         |    /        +--------+        \    |         |
+     |         v   v          /  |              v   v         |
+     |     +----------+      /   |           +----------+     |
+     |     |   half   |     /    |           |   half   |     |
+     |     |  closed  |    /     | send R    |  closed  |     |
+     |     | (remote) |   /      |           | (local)  |     |
+     |     +----------+  .       |           +----------+     |
```
Scottmitch: nit: `,` seems to be the angled character in this direction...did this not look pretty here?

draft-ietf-httpbis-http2.xml
```diff
@@ -1904,12 +1906,13 @@ HTTP2-Settings    = token68
         </t>
 
         <t>
-          The RST_STREAM frame fully terminates the referenced stream and causes it to enter the
-          closed state. After receiving a RST_STREAM on a stream, the receiver MUST NOT send
-          additional frames for that stream, with the exception of <x:ref>PRIORITY</x:ref>. However,
-          after sending the RST_STREAM, the sending endpoint MUST be prepared to receive and process
-          additional frames sent on the stream that might have been sent by the peer prior to the
-          arrival of the RST_STREAM.
+          Receiving a RST_STREAM frame fully terminates the stream and causes it to enter the
```
Scottmitch: RST_STREAM ->  <x:ref>RST_STREAM</x:ref>?

Scottmitch: This applies to the whole paragraph.

draft-ietf-httpbis-http2.xml
```diff
@@ -1904,12 +1906,13 @@ HTTP2-Settings    = token68
...
-          closed state. After receiving a RST_STREAM on a stream, the receiver MUST NOT send
-          additional frames for that stream, with the exception of <x:ref>PRIORITY</x:ref>. However,
-          after sending the RST_STREAM, the sending endpoint MUST be prepared to receive and process
-          additional frames sent on the stream that might have been sent by the peer prior to the
-          arrival of the RST_STREAM.
+          Receiving a RST_STREAM frame fully terminates the stream and causes it to enter the 
+          "closed" state.  After receiving a RST_STREAM on a stream, the receiver MUST NOT send
+          additional frames for that stream, with the exception of <x:ref>PRIORITY</x:ref>. 
+          Sending a RST_STREAM causes the stream to enter the "reset" state and wait for a RST_STREAM
+          in response. After sending the RST_STREAM, the sending endpoint MUST be prepared to receive
```
Scottmitch: `in response` -> `from the remote peer`?

Scottmitch: Humm not ideal in general but the variety of terminology for this type of situation seems to be:
1. `in response`
2. `remote peer`
3. some form of `endpoint`

I'm not a big fan of the `response` terminology because it has meaning within the HTTP protocol.

draft-ietf-httpbis-http2.xml
gregw: I don't think the xref is needed as we are in the referenced paragraph.  It would just be a self reference

Scottmitch: Fair enough.  Then I think it needs to be removed in the paragraph below.  I'll create a new comment.

draft-ietf-httpbis-http2.xml
```diff
@@ -1926,6 +1929,13 @@ HTTP2-Settings    = token68
           <x:ref>PROTOCOL_ERROR</x:ref>.
         </t>
         <t>
+          An endpoint MUST send a <x:ref>RST_STREAM</x:ref> with <x:ref>NO_ERROR</x:ref> in
```
Scottmitch: `<x:ref>RST_STREAM</x:ref>` -> RST_STREAM

martinthomson:
HTTP/2 is approved and going to the RFC editor.

reschke:
Nit: https://datatracker.ietf.org/doc/draft-ietf-httpbis-http2/history/ doesn't show it as approved.

PR #703, opened 2015-01-23
Addressing concerns about the use of an ALPN identifier

martinthomson:
... for a non-ALPN/non-TLS purpose.

Raised by Joe Salowey in reviewing the "h2c" ALPN registration.

### DIFF
--- a/draft-ietf-httpbis-http2.xml
+++ b/draft-ietf-httpbis-http2.xml
@@ -467,12 +467,14 @@ HTTP2-Settings    = token68
 
       <section anchor="discover-https" title="Starting HTTP/2 for &quot;https&quot; URIs">
         <t>
-          A client that makes a request to an "https" URI uses <xref target="TLS12">TLS</xref>
-          with the <xref target="TLS-ALPN">application layer protocol negotiation extension</xref>.
+          A client that makes a request to an "https" URI uses <xref target="TLS12">TLS</xref> with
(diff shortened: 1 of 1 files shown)


jsalowey:
Looks good.

Thanks,

Joe

draft-ietf-httpbis-http2.xml
```diff
@@ -467,12 +467,14 @@ HTTP2-Settings    = token68
...
+          A client that makes a request to an "https" URI uses <xref target="TLS12">TLS</xref> with
+          the <xref target="TLS-ALPN">application layer protocol negotiation (ALPN)
+          extension</xref>.
         </t>
         <t>
-          HTTP/2 over TLS uses the "h2" application token.  The "h2c" token MUST NOT be sent by a
-          client or selected by a server.
+          HTTP/2 over TLS uses the "h2" application identifier.  The "h2c" token MUST NOT be sent by
+          a client or selected by a server; the "h2c" token is reserved from the ALPN identifier
+          space, but describes a protocol that does not use TLS.
```
hruellan: I'd simplify with:
the "h2c" token is used to describe a protocol that does not use TLS.

martinthomson: I think that the bit that elides is critical to avoid the sorts of concerns that Joe (and others) raised.

hruellan: I missed that.
I'd probably move this to 3.1. See my proposal in #710.

PR #704, opened 2015-01-23
Expands never indexed literals guidance.

hruellan:
Expands section 7.1.3 to describes characteristics of header needing the protection offered by the never indexed literal representation.

### DIFF
--- a/draft-ietf-httpbis-header-compression.xml
+++ b/draft-ietf-httpbis-header-compression.xml
@@ -1251,14 +1251,31 @@ else
                         representation MUST be used.
                     </t>
                     <t>
-                        The decision on whether a header field is sensitive or
-                        not is highly dependent on the context. As a generic
-                        guidance, header fields used for conveying highly valued
-                        information, such as the Authorization or Cookie header
-                        fields, can be considered to be on the more sensitive
-                        side.  In addition, a header field with a short value
-                        has potentially a smaller entropy and can be more at
-                        risk.
+                        The choice to use a never indexed literal representation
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-header-compression.xml
```diff
@@ -1227,24 +1227,38 @@ else
...
+                    </t>
+                    <t>
+                        The decision on whether a header field is sensitive or
+                        not is highly dependent on the context. As a generic
+                        guidance, header fields used for conveying highly valued
+                        information, such as the Authorization or Cookie header
+                        fields, can be considered to be on the more sensitive
+                        side.  In addition, a header field with a short value
+                        has potentially a smaller entropy and can be more at
+                        risk.
```
martinthomson: How about:
> The choice to use a never indexed literal representation for a header field depends on several factors.  Since HPACK compresses based on the entirety of a value, short or low-entropy values are more readily recovered by an adversary.  Therefore, an encoder might choose to disable compression for values with low entropy.  Encoders might also choose to disable compression for header fields that are considered to be highly valuable or sensitive to recovery, such as the Cookie or Authorization header fields.

> An encoder might prefer compression for values that have little or no value if they were exposed.  For instance, a User-Agent header field does not commonly vary between requests and is sent to any server.  In that case, confirmation that a particular User-Agent value has been used provides little value.

hruellan: Are you proposing to drop the "An intermediary MUST NOT..." paragraph?

martinthomson:
I think that this looks good now.

PR #710, opened 2015-01-27
Alternative to #703

hruellan:
Moves considerations of defining an ALPN token for a non-TLS-based protocol to 3.1.
Changes "token" to "protocol identifier" in 3.3.

### DIFF
--- a/draft-ietf-httpbis-http2.xml
+++ b/draft-ietf-httpbis-http2.xml
@@ -306,9 +306,8 @@
                 HTTP/2 over TCP is identified.
               </t>
               <t>
-                The "h2c" string is serialized into an ALPN protocol identifier as the three octet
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2.xml
```diff
@@ -305,6 +305,11 @@
                 This identifier is used in the HTTP/1.1 Upgrade header field and in any place where
                 HTTP/2 over TCP is identified.
               </t>
+              <t>
+                The "h2c" string is serialized into an ALPN protocol identifier as the three octet
+                sequence: 0x68, 0x32, 0x63. It is reserved from the ALPN identifier space, but
+                describes a protocol that does not use TLS.
+              </t>
```
martinthomson: I think that the second sentence is all that we need here.  We don't need to describe h2c as a series of octets (in Upgrade, it appears literally, and we still have the IANA section).

PR #712, opened 2015-02-06
Uses decoder perspective for maximum table size changes.

hruellan:


draft-ietf-httpbis-header-compression.xml
```diff
@@ -556,11 +553,11 @@
...
                 <t>
                     Multiple updates to the maximum table size can occur between
-                    the sending of two header blocks.  In the case that this
+                    the transmission of two header blocks.  In the case that this
                     size is changed more than once in this interval, the
                     smallest maximum table size that occurs in that interval
-                    MUST be sent in an encoding context update.  The final
-                    maximum size is always sent, resulting in at most two
+                    MUST be signalled in an encoding context update.  The final
+                    maximum size is always signalled, resulting in at most two
```
martinthomson: In American English, it's "signaled". Weird, but they are like that sometimes...

PR #714, opened 2015-02-10
Better explanation for static header table.

hruellan:


martinthomson:
I'll leave this open as a reminder that we need an RFC editor's note for it.  Text proposed on-list.

hruellan:
Handled by 2985dbe.

PR #718, opened 2015-02-13
H2 17 edburns editorial

edburns:


martinthomson:
This one seems to add to the last one.  I don't think the changes in the new commits are necessary though.

PR #719, opened 2015-02-19
Fix some typos

leekchan:
Hi.

I fixed some typos.


mnot:
Thanks for that, but the portion of the compression spec you corrected is going to be removed when published, and the XSLT is copied from elsewhere, so updating it here won't help.

reschke:
I have fixed the upstream XSLT, and copied the latest version over here.

mnot:
Thanks, Julian!

PR #720, opened 2015-02-19
Clear up HPACK Section 4.1 "Calculating Table Size"

jrudolph:
It was not quite clear what exactly was referenced with "as defined in Section 5.2". 
In section 5.2, "String length" is defined as the "number of octets used to encode the string". 
This, however, would contradict the (now deleted) second paragraph which explicitly states that the 
length is to be calculated "without any Huffman encoding applied".

Therefore, the reference was deleted and both paragraphs consolidated into one.

### DIFF
--- a/draft-ietf-httpbis-header-compression.xml
+++ b/draft-ietf-httpbis-header-compression.xml
@@ -503,14 +503,10 @@
                     entries.
                 </t>
                 <t>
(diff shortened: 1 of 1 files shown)


hruellan:
Your suggestion would improve the HPACK specification.
Unfortunately, HPACK has been approved and is being published.

PR #724, opened 2015-03-16
Adding protection from padding for the fields in HEADERS

### LINKED ISSUE #723: Fix padding descriptions
This wasn't obvious, but all descriptions of padding reference DATA:

> If the length of the padding is the length of the frame payload or greater, the recipient MUST treat this as a connection error (Section 5.4.1) of type PROTOCOL_ERROR. 

However, this is subtly wrong, as @daurnimator notes.  The actual text should be:

> If the length of the padding exceeds the length of the data (for HEADERS, the header block), then the recipient MUST ...

martinthomson:
Closes #723

### DIFF
--- a/draft-ietf-httpbis-http2.xml
+++ b/draft-ietf-httpbis-http2.xml
@@ -1817,7 +1817,8 @@ HTTP2-Settings    = token68
 
         <t>
           The HEADERS frame can include padding.  Padding fields and flags are identical to those
-          defined for <xref target="DATA">DATA frames</xref>.
+          defined for <xref target="DATA">DATA frames</xref>.  Padding that exceeds the size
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2.xml
```diff
@@ -1817,7 +1817,8 @@ HTTP2-Settings    = token68
 
         <t>
           The HEADERS frame can include padding.  Padding fields and flags are identical to those
-          defined for <xref target="DATA">DATA frames</xref>.
+          defined for <xref target="DATA">DATA frames</xref>.  Padding that exceeds the size
+          remaining to the header block fragment MUST be treated as a <x:ref>PROTOCOL_ERROR</x:ref>.
```
daurnimator: s/to/in/

hruellan:
LGTM

martinthomson:
Merged by RFC editor.

PR #725, opened 2015-03-16
'at any time' for PRIORITY was confusing people

martinthomson:
This was confusing a few people who weren't clear if the prohibition on frame sending for CONTINUATION trumped this or not.  This isn't a perfect fix, but it should be enough.


hruellan:
LGTM

martinthomson:
Merged by RFC editor.

PR #732, opened 2015-03-22
Fixing bug with GOAWAY and graceful shutdown.

### LINKED ISSUE #731: Some clarifications regarding GOAWAY.
We are working on a HTTP/2 library and we are a bit confused and we would like to get it right.

Is it correct that after a sender sent a `GOAWAY` frame with some `last stream identifier` to a receiver, the sender may still open new streams, even with stream identifiers higher than the `last stream identifier` sent in the `GOAWAY` frame? That is the `GOAWAY` frame and the `last stream identifier` only limit the ability of the receiver?

If this is correct, and I believe it is, then I find the following sentence in the first paragraph at `6.8 GOAWAY` confusing.

> Once sent, the sender will ignore frames sent on any new streams with identifiers higher than the included last stream identifier.

In my opinion it should say something of the like 

> Once sent, the sender will ignore frames sent on any new streams **created by the receiver** with identifiers higher than the included last stream identifier.

Thanks.

martinthomson:
Closes #731.

### DIFF
--- a/draft-ietf-httpbis-http2.xml
+++ b/draft-ietf-httpbis-http2.xml
@@ -2449,14 +2449,14 @@ HTTP2-Settings    = token68
         </t>
 
         <t>
-          After sending a GOAWAY frame, the sender can discard frames for streams with identifiers
-          higher than the identified last stream.  However, any frames that alter connection state
-          cannot be completely ignored.  For instance, <x:ref>HEADERS</x:ref>,
-          <x:ref>PUSH_PROMISE</x:ref> and <x:ref>CONTINUATION</x:ref> frames MUST be minimally
-          processed to ensure the state maintained for header compression is consistent (see <xref
-          target="HeaderBlock"/>); similarly DATA frames MUST be counted toward the connection flow
-          control window.  Failure to process these frames can cause flow control or header
-          compression state to become unsynchronized.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2.xml
```diff
@@ -2337,9 +2337,10 @@ HTTP2-Settings    = token68
         <t>
           The GOAWAY frame (type=0x7) informs the remote peer to stop creating streams on this
           connection.  GOAWAY can be sent by either the client or the server.  Once sent, the sender
-          will ignore frames sent on any new streams with identifiers higher than the included last
-          stream identifier.  Receivers of a GOAWAY frame MUST NOT open additional streams on the
-          connection, although a new connection can be established for new streams.
+          will ignore frames sent on streams initiated by the reciever if the stream has an
```
buchgr: @martinthomson s/reciever/receiver/

buchgr:
@martinthomson thanks!

A few paragraphs further down it also says

> After sending a GOAWAY frame, the sender can discard frames for streams with identifiers higher than the identified last stream.

For clarity, this should probably also say for `streams initiated by the receiver` here.

martinthomson:
I think that we've settled on #733

PR #733, opened 2015-03-22
Goaway fix v2

### LINKED ISSUE #731: Some clarifications regarding GOAWAY.

martinthomson:
After some sleep, I've read through the section and noticed a few other minor inconsistencies.  This alters the structure slightly, as well as replacing the first sentence with a more accurate one.

Supercedes #732.  Closes #731, #732.

### DIFF
--- a/draft-ietf-httpbis-http2.xml
+++ b/draft-ietf-httpbis-http2.xml
@@ -2449,14 +2449,14 @@ HTTP2-Settings    = token68
         </t>
 
         <t>
-          After sending a GOAWAY frame, the sender can discard frames for streams with identifiers
-          higher than the identified last stream.  However, any frames that alter connection state
-          cannot be completely ignored.  For instance, <x:ref>HEADERS</x:ref>,
-          <x:ref>PUSH_PROMISE</x:ref> and <x:ref>CONTINUATION</x:ref> frames MUST be minimally
-          processed to ensure the state maintained for header compression is consistent (see <xref
-          target="HeaderBlock"/>); similarly DATA frames MUST be counted toward the connection flow
-          control window.  Failure to process these frames can cause flow control or header
-          compression state to become unsynchronized.
+          After sending a GOAWAY frame, the sender can discard frames for streams initiated by the
+          receiver with identifiers higher than the identified last stream.  However, any frames
+          that alter connection state cannot be completely ignored.  For instance,
+          <x:ref>HEADERS</x:ref>, <x:ref>PUSH_PROMISE</x:ref> and <x:ref>CONTINUATION</x:ref> frames
+          MUST be minimally processed to ensure the state maintained for header compression is
+          consistent (see <xref target="HeaderBlock"/>); similarly DATA frames MUST be counted
+          toward the connection flow control window.  Failure to process these frames can cause flow
+          control or header compression state to become unsynchronized.
         </t>
 
         <t>
--- a/draft-ietf-httpbis-http2.xml
+++ b/draft-ietf-httpbis-http2.xml
@@ -2438,14 +2438,15 @@ HTTP2-Settings    = token68
         </t>
         <t>
           A client that is unable to retry requests loses all requests that are in flight when the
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2.xml
```diff
@@ -2443,14 +2449,14 @@ HTTP2-Settings    = token68
...
-          After sending a GOAWAY frame, the sender can discard frames for streams with identifiers
-          higher than the identified last stream.  However, any frames that alter connection state
-          cannot be completely ignored.  For instance, <x:ref>HEADERS</x:ref>,
-          <x:ref>PUSH_PROMISE</x:ref> and <x:ref>CONTINUATION</x:ref> frames MUST be minimally
-          processed to ensure the state maintained for header compression is consistent (see <xref
-          target="HeaderBlock"/>); similarly DATA frames MUST be counted toward the connection flow
-          control window.  Failure to process these frames can cause flow control or header
-          compression state to become unsynchronized.
+          After sending a GOAWAY frame, the sender can discard frames for streams initiated by the
+          receiver with identifiers higher than the identified last stream.  However, any frames
```
martinthomson: Note: only change is in the first two lines.

buchgr:
+1

draft-ietf-httpbis-http2.xml
Scottmitch: `with identifiers` -> `with an identifier`?

draft-ietf-httpbis-http2.xml
```diff
@@ -2371,6 +2372,11 @@ HTTP2-Settings    = token68
           An endpoint might choose to close a connection without sending GOAWAY for misbehaving
           peers.
         </t>
+        <t>
+          A GOAWAY frame might not immediately precede closing of the connection; a receiver of a
```
Scottmitch: `might not` -> `may not`?

draft-ietf-httpbis-http2.xml
```diff
@@ -2354,6 +2348,13 @@ HTTP2-Settings    = token68
           the identified stream is the highest numbered stream initiated by the client.
         </t>
         <t>
+          Once sent, the sender will ignore frames sent on streams initiated by the receiver if the
```
Scottmitch: `Once sent, the sender will ignore` is it possible to use a familiar term instead of `will` (i.e. `MUST`, `SHOULD`, `may`, etc...)?

draft-ietf-httpbis-http2.xml
martinthomson: either works, I think

draft-ietf-httpbis-http2.xml
martinthomson: which might be interpreted (incorrectly) as `MAY not`, which is a common gaffe that @reschke won't allow

draft-ietf-httpbis-http2.xml
martinthomson: This is simply moved text.  I think that a statement of fact rather than requirement is best.

draft-ietf-httpbis-http2.xml
Scottmitch: +1

hruellan:
LGTM

martinthomson:
Merged by RFC editor.

PR #735, opened 2015-03-24
Reformating cipher black list

### LINKED ISSUE #722: Reformat cipher suite list
Just so I don't forget, make this a proper list rather than a comma-separated blob (the latter is hard to scan).

martinthomson:
Closes #722


hruellan:
LGTM!

martinthomson:
Merged by RFC editor.

PR #736, opened 2015-03-24
Noting effect of settings on flow control more precisely

### LINKED ISSUE #727: initial flow control window size state clarification
[section 6.9.2](https://tools.ietf.org/html/draft-ietf-httpbis-http2-17#section-6.9.2) has the following text:

> A SETTINGS frame can alter the initial flow control window size for
   all streams in the "open" or "half closed (remote)" state.  When the
   value of SETTINGS_INITIAL_WINDOW_SIZE changes, a receiver MUST adjust
   the size of all stream flow control windows that it maintains by the
   difference between the new value and the old value.

Draft 17 added the changed `current streams` to `streams in the "open" or "half closed (remote)" state`.  Can someone clarify why would streams in other states be excluded?

martinthomson:
Closes #727

### DIFF
--- a/draft-ietf-httpbis-http2.xml
+++ b/draft-ietf-httpbis-http2.xml
@@ -2601,11 +2601,12 @@ HTTP2-Settings    = token68
             received.
           </t>
           <t>
-            A <x:ref>SETTINGS</x:ref> frame can alter the initial flow control window size for all
-            streams in the "open" or "half closed (remote)" state. When the value of
-            <x:ref>SETTINGS_INITIAL_WINDOW_SIZE</x:ref> changes, a receiver MUST adjust the size of
-            all stream flow control windows that it maintains by the difference between the new
-            value and the old value.
+            In addition to changing the flow control window for future streams, a
+            <x:ref>SETTINGS</x:ref> frame can alter the initial flow control window size for streams
+            with active flow control windows (that is, streams in the "open" or "half closed
+            (remote)" state).  When the value of <x:ref>SETTINGS_INITIAL_WINDOW_SIZE</x:ref>
+            changes, a receiver MUST adjust the size of all stream flow control windows that it
+            maintains by the difference between the new value and the old value.
           </t>
           <t>
             A change to <x:ref>SETTINGS_INITIAL_WINDOW_SIZE</x:ref> can cause the available space in
--- a/draft-ietf-httpbis-http2.xml
+++ b/draft-ietf-httpbis-http2.xml
@@ -2601,7 +2601,7 @@ HTTP2-Settings    = token68
             received.
           </t>
           <t>
-            In addition to changing the flow control window for future streams, a
+            In addition to changing the flow control window for streams that are not yet active, a
             <x:ref>SETTINGS</x:ref> frame can alter the initial flow control window size for streams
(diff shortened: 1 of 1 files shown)


Scottmitch:
Thanks for the PR @martinthomson!  Should we move the issue discussion (https://github.com/http2/http2-spec/issues/727#issuecomment-85060068) to this PR?

hruellan:
LGTM

martinthomson:
@Scottmitch, if you have comments, please let me know ASAP.

Scottmitch:
Below is from the issue (https://github.com/http2/http2-spec/issues/727#issuecomment-85060068):

I agree that it doesn't need to be specifically called out that the window can be treated as a resource, and that this is an implantation detail.  However calling out the stream states when referencing frames that impact the flow control window may be overly restrictive.  For example the `"open" or "half closed (remote)"` are not the only states in which a stream may be in the priority tree.

Here are what I believe to be the relevant portions of the specification which relate to flow control window and stream state.

1. [section 6.9](https://tools.ietf.org/html/draft-ietf-httpbis-http2-17#section-6.9) seems to be very careful about stream state and doesn't necessarily limit to just `"open" or "half closed (remote)"`
2. [section 5.1](https://tools.ietf.org/html/draft-ietf-httpbis-http2-17#section-5.1) is much more restrictive and declares that `WINDOW_UPDATE` is not allowed in `idle`, or `closed`. For `closed` there is special mention that these frames must be ignored, or may be an error (if it arrives `a significant time after sending END_STREAM`).
3. If the clarifications in this issue are applied then it seems like the `SETTINGS` impact on the flow control windows is further limited to `"open" or "half closed (remote)"`.

draft-ietf-httpbis-http2.xml
```diff
@@ -2601,11 +2601,12 @@ HTTP2-Settings    = token68
             received.
           </t>
           <t>
-            A <x:ref>SETTINGS</x:ref> frame can alter the initial flow control window size for all
-            streams in the "open" or "half closed (remote)" state. When the value of
-            <x:ref>SETTINGS_INITIAL_WINDOW_SIZE</x:ref> changes, a receiver MUST adjust the size of
-            all stream flow control windows that it maintains by the difference between the new
-            value and the old value.
+            In addition to changing the flow control window for future streams, a
```
Scottmitch: I know I provided the `future streams` terminology in my issue report, and I understand its use is helpful to distinguish that per stream windows are not required to be maintained for streams in the IDLE state.  However the `future streams` seems to be a new expression.  I'm wondering if it makes sense to re-word to be more consistent and avoid introducing new expressions?  What about something like:

`A SETTINGS frame can alter the initial flow control window size for streams with active flow control windows (i.e. "open", and "half closed (remote)) and when streams transition out of the "idle" state and active flow control windows are allocated.`

Scottmitch: Note that I updated the alternative suggestion text above since the original post.

hruellan: What about changing the first line to:
> In addition to changing the initial flow control window for streams not yet opened, a

martinthomson:
Merged by RFC editor.

PR #737, opened 2015-03-24
PUSH_PROMISE sending conditions

### LINKED ISSUE #726: push_promise frame state clarification
[section-6.6](https://tools.ietf.org/html/draft-ietf-httpbis-http2-17#section-6.6) has the following clarification added in draft 17:
> PUSH_PROMISE frames MUST be associated with a peer-initiated stream
   that is in either the "open" or "half closed (remote)" state.

Should this be `half closed (local)` instead of `half closed (remote)`?

If not I think we have an inconsistency with [section 5.1](https://tools.ietf.org/html/draft-ietf-httpbis-http2-17#section-5.1) which has the following for the `half closed (remote)` section:
> half closed (remote):
      A stream that is "half closed (remote)" is no longer being used by
      the peer to send frames.  In this state, an endpoint is no longer
      obligated to maintain a receiver flow control window.
      
>If an endpoint receives additional frames for a stream that is in
      this state, other than WINDOW_UPDATE, PRIORITY or RST_STREAM, it
      MUST respond with a stream error (Section 5.4.2) of type
      STREAM_CLOSED.

martinthomson:
This makes it clearer that they can't be sent otherwise.

Closes #726

### DIFF
--- a/draft-ietf-httpbis-http2.xml
+++ b/draft-ietf-httpbis-http2.xml
@@ -2222,8 +2222,8 @@ HTTP2-Settings    = token68
         </t>
 
         <t>
-          PUSH_PROMISE frames MUST be associated with a peer-initiated stream that is in either the
-          "open" or "half closed (remote)" state. The stream identifier of a PUSH_PROMISE frame
+          PUSH_PROMISE frames MUST only be sent on with a peer-initiated stream that is in either
+          the "open" or "half closed (remote)" state. The stream identifier of a PUSH_PROMISE frame
           indicates the stream it is associated with.  If the stream identifier field specifies the
           value 0x0, a recipient MUST respond with a <xref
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2.xml
```diff
@@ -2222,8 +2222,8 @@ HTTP2-Settings    = token68
         </t>
 
         <t>
-          PUSH_PROMISE frames MUST be associated with a peer-initiated stream that is in either the
-          "open" or "half closed (remote)" state. The stream identifier of a PUSH_PROMISE frame
+          PUSH_PROMISE frames MUST only be sent on with a peer-initiated stream that is in either
```
hruellan: s/sent on with/sent on/ ?

hruellan:
LGTM

martinthomson:
Merged by RFC editor.

PR #742, opened 2015-04-07
Fix "an" typos

brendanashworth:
Just a quick look through. Thanks!


martinthomson:
The RFC editor is [working on this now](http://www.rfc-editor.org/cluster_info.php?cid=C243).  We'll let them fix these.

PR #753, opened 2016-05-06
Change upgrade title to say HTTP/1.1

phroggyy:
I'm not sure if the intention was to have _Upgrading to HTTP/2_ or _Upgrading from HTTP/1.1_, but the current title is misleading and hence demands a change.


mnot:
This section is about upgrading from HTTP/2 to something else.

PR #763, opened 2017-06-27
Fix markdown formatting in README

Notoh:
There is no space between the two #'s for the proper markdown header.


mnot:
Fixed in source, thanks!

PR #795, opened 2020-12-22
Remove priority

### LINKED ISSUE #773: Consider (partial) removal of priority signaling
This probably needs a skeleton left over so that we can avoid interoperability problems that might arise from interactions with peers that still have them. There are some MUST-level requirements around parsing that would need to be identified and marked.

See https://github.com/martinthomson/http2v2/issues/3 for more discussion.

martinthomson:
This doesn't completely remove it, but it only keeps those pieces
necessary for interoperability with existing implementations: the rules
for sending PRIORITY frames in different stream states, the format of
the HEADERS and PRIORITY frames, and the error handling.

It replaces the exposition about how priority works with a deprecation
notice and the little advice that we currently understand to be
accurate.

Closes #773.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -1200,7 +1200,7 @@
         <t>
           In a multiplexed protocol like HTTP/2, prioritizing allocation of bandwidth and
           computation resources to streams can be critical to attaining good performance.  A poor
-          prioritization scheme can result in HTTP/2 providing poor performance.  With no parallism
+          prioritization scheme can result in HTTP/2 providing poor performance.  With no parallelism
           at the TCP layer, performance could be significantly worse than HTTP/1.1.
         </t>
         <t>
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -888,9 +888,10 @@
                   </li>
             </ul>
             <t>
-                An endpoint MAY send a <xref target="PRIORITY" format="none">PRIORITY</xref> frame in this state.  An endpoint MUST NOT send any type of frame other than
-                <xref target="RST_STREAM" format="none">RST_STREAM</xref>, <xref target="WINDOW_UPDATE" format="none">WINDOW_UPDATE</xref>, or <xref target="PRIORITY" format="none">PRIORITY</xref>
-                in this state.
+                An endpoint MUST NOT send any type of frame other than <xref target="RST_STREAM"
+                format="none">RST_STREAM</xref>, <xref target="WINDOW_UPDATE"
+                format="none">WINDOW_UPDATE</xref>, or <xref target="PRIORITY"
+                format="none">PRIORITY</xref> in this state.
             </t>
             <t>
                 Receiving any type of frame other than <xref target="HEADERS" format="none">HEADERS</xref>,
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -1629,7 +1629,7 @@
         <t>
           The HEADERS frame can include padding.  Padding fields and flags are identical to those
           defined for <xref target="DATA">DATA frames</xref>.  Padding that exceeds the size
-          remaining for the header block fragment MUST be treated as a
+          remaining for the field block fragment MUST be treated as a
           <xref target="PROTOCOL_ERROR" format="none">PROTOCOL_ERROR</xref>.
         </t>
       </section>
@@ -1679,7 +1679,7 @@
           Sending or receiving a PRIORITY frame does not affect the state of any stream (<xref
           target="StreamStates"/>).  The PRIORITY frame can be sent on a stream in any state,
           including "idle" or "closed".  A PRIORITY frame cannot be sent between consecutive frames
-          that comprise a single <xref target="HeaderBlock">header block</xref>.
+          that comprise a single <xref target="FieldBlock">field block</xref>.
         </t>
         <t>
           A PRIORITY frame with a length other than 5 octets MUST be treated as a <xref
@@ -3149,7 +3149,7 @@
             image links ensures that the client is able to see that a resource will be pushed
             before discovering embedded links. Similarly, if the server pushes responses referenced by the field block
             (for instance, in Link header fields), sending a <xref target="PUSH_PROMISE" format="none">PUSH_PROMISE</xref> before sending the
-            header block ensures that clients do not request those resources.
+            header ensures that clients do not request those resources.
           </t>
           <t><xref target="PUSH_PROMISE" format="none">PUSH_PROMISE</xref> frames MUST NOT be sent by the client.
           </t>
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -1636,8 +1636,8 @@
       <section anchor="PRIORITY">
         <name>PRIORITY</name>
         <t>
-          The PRIORITY frame (type=0x2) is deprecated; see <xref target="PriorityHere"/> A PRIORITY
-          frame can be sent in any stream state, including idle or closed streams.
+          The PRIORITY frame (type=0x2) is deprecated; see <xref target="PriorityHere"/>.  A
+          PRIORITY frame can be sent in any stream state, including idle or closed streams.
         </t>
         <figure anchor="PRIORITYFramePayload">
           <name>PRIORITY Frame Payload</name>
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -1249,7 +1249,7 @@
           <t>
             Though the priority signaling from RFC 7540 was not widely adopted, the information it
             provides can still be useful in the absence of better information.  Endpoints that
-            receive priority signals <xref target="HEADERS" format="none">HEADERS</xref> or <xref
+            receive priority signals in <xref target="HEADERS" format="none">HEADERS</xref> or <xref
             target="PRIORITY" format="none">PRIORITY</xref> frames can benefit from applying that
             information.  In particular, implementations that consume these signals would not
             benefit from discarding these priority signals in the absence of alternatives.
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -367,10 +367,10 @@
         </t>
         <t>
           The HTTP/1.1 request that is sent prior to upgrade is assigned a stream identifier of 1
-          (see <xref target="StreamIdentifiers"/>) with <xref target="pri-default">default priority
-          values</xref>.  Stream 1 is implicitly "half-closed" from the client toward the server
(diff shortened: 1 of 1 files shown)


draft-thomson-httpbis-http2bis.xml
```diff
@@ -1737,20 +1577,12 @@
...
-          stream</xref>.
-        </t>
       </section>
       <section anchor="PRIORITY">
         <name>PRIORITY</name>
         <t>
-          The PRIORITY frame (type=0x2) specifies the <xref target="StreamPriority">sender-advised
-          priority of a stream</xref>.  It can be sent in any stream state, including idle or closed
-          streams.
+          The PRIORITY frame (type=0x2) is deprecated; see <xref target="PriorityHere"/> A PRIORITY
```
Lukasa, line 1584: Is there a period missing here?

Lukasa (approved):
LGTM.

draft-thomson-httpbis-http2bis.xml
```diff
@@ -1167,231 +1162,77 @@
...
-          be used to select streams for transmitting frames when there is limited capacity for
-          sending.
-        </t>
+        <name>Prioritization</name>
         <t>
-          Streams can be prioritized by marking them as dependent on the completion of other streams
-          (<xref target="pri-depend"/>).  Each dependency is assigned a relative weight, a number
-          that is used to determine the relative proportion of available resources that are assigned
-          to streams dependent on the same stream.
+          In a multiplexed protocol like HTTP/2, prioritizing allocation of resources to streams can
```
rmarx, line 1169: This is imo somewhat unclear language: what exactly are the "resources" here? When talking about web page construction in other contexts, a "resource" is often a file (say main.js), while I feel here resources is more akin to bandwidth (as files are implicitly linked to individual streams?)

draft-thomson-httpbis-http2bis.xml
```diff
@@ -1167,231 +1162,77 @@
...
-          turns out that they are not useful, or actively harmful, implementations will be requested
-          to avoid creating stream dependencies.
-        -->
         <t>
-          Explicitly setting the priority for a stream is input to a prioritization process.  It
-          does not guarantee any particular processing or transmission order for the stream relative
-          to any other stream.  An endpoint cannot force a peer to process concurrent streams in a
-          particular order using priority.  Expressing priority is therefore only a suggestion.
+          A good prioritization scheme benefits from the application of contextual knowledge such as
+          the content of resources, how resources are interrelated, and how those resources will be
```
rmarx, line 1176:
```suggestion
          the content and size of resources, how resources are interrelated, and how those resources will be
```

draft-thomson-httpbis-http2bis.xml
```diff
@@ -1167,231 +1162,77 @@
...
-            of the resources allocated to stream C.
-          </t>
-        </section>
-        <section anchor="reprioritize">
-          <name>Reprioritization</name>
-          <t>
-            Stream priorities are changed using the <xref target="PRIORITY" format="none">PRIORITY</xref> frame.  Setting a
-            dependency causes a stream to become dependent on the identified parent stream.
+            HTTP/2 included a rich system for signaling priority of requests.  However, this system
+            proved to be complex and it was not widely implemented.
```
rmarx, line 1185: Not sure I agree with "not widely implemented"... most H2 implementations have a (basic) version of this. 

"and it's full flexibility was rarely employed" or "and it was rarely correctly implemented" or similar would fit the bill better. (or it can be left out, as this is also said in other terms below)

draft-thomson-httpbis-http2bis.xml
```diff
@@ -1167,231 +1162,77 @@
...
           </t>
           <t>
-            The amount of additional state an endpoint maintains for prioritization could be
-            dependent on load; under high load, prioritization state can be discarded to limit
-            resource commitments.  In extreme cases, an endpoint could even discard prioritization
-            state for active or reserved streams. If a limit is applied, endpoints SHOULD maintain
-            state for at least as many streams as allowed by their setting for
-            <xref target="SETTINGS_MAX_CONCURRENT_STREAMS" format="none">SETTINGS_MAX_CONCURRENT_STREAMS</xref>.  Implementations SHOULD also attempt to
-            retain state for streams that are in active use in the priority tree.
+            Signaling priority information could still be necessary to attain good performance.
```
rmarx, line 1214: Above, it is said that good prioritization can be critical and poor prioritization can lead to performance degradation, which I agree with. As such "could still be necessary" is too weak of a statement. "is still an important aspect to attaining good HTTP/2 performance" would be better.

draft-thomson-httpbis-http2bis.xml
```diff
@@ -1167,231 +1162,77 @@
...
+            alternative scheme, such as <xref target="I-D.ietf-httpbis-priority"/>.
           </t>
           <t>
-            If it has retained enough state to do so, an endpoint receiving a
-            <xref target="PRIORITY" format="none">PRIORITY</xref> frame that changes the priority of a
-            closed stream SHOULD alter the dependencies of the streams that depend on it.
+            Though the priority signaling from RFC 7540 was not widely adopted, the information it
+            provides can still be useful in the absence of better information.  Endpoints that
+            receive <xref target="HEADERS" format="none">HEADERS</xref> or <xref target="PRIORITY"
+            format="none">PRIORITY</xref> frames can benefit from applying any priority that is
```
rmarx, line 1222:
```suggestion
            format="none">PRIORITY</xref> frames can benefit from applying any priority signals that are
```

draft-thomson-httpbis-http2bis.xml
```diff
@@ -1167,231 +1162,77 @@
...
+            Where signaling priority information is important, endpoints are encouraged to use an
+            alternative scheme, such as <xref target="I-D.ietf-httpbis-priority"/>.
           </t>
           <t>
-            If it has retained enough state to do so, an endpoint receiving a
-            <xref target="PRIORITY" format="none">PRIORITY</xref> frame that changes the priority of a
-            closed stream SHOULD alter the dependencies of the streams that depend on it.
+            Though the priority signaling from RFC 7540 was not widely adopted, the information it
+            provides can still be useful in the absence of better information.  Endpoints that
+            receive <xref target="HEADERS" format="none">HEADERS</xref> or <xref target="PRIORITY"
```
rmarx, line 1221: Seeing as the extensible priorities draft also utilizes HTTP headers to communicate its signals, I wonder if this mention of HEADERS frames is too ambiguous for uninitiated readers here? Maybe something like "HEADERS (as defined in RFC 7540)" ?

draft-thomson-httpbis-http2bis.xml
```diff
@@ -1167,231 +1162,77 @@
...
-        <section anchor="pri-default">
-          <name>Default Priorities</name>
           <t>
-            All streams are initially assigned a non-exclusive dependency on stream 0x0.  <xref target="PushResources">Pushed streams</xref> initially depend on their associated
-            stream.  In both cases, streams are assigned a default weight of 16.
+            Servers SHOULD use other contextual information in determining priority of requests in
+            the absence of any priority signals.  In particular, servers might need to decide
+            whether to interpret the complete absence of signals as an indication that the client
+            has not implemented the feature or that the client that has deliberately chosen to rely
+            on the defaults described in <xref target="RFC7540" section="5.3.5"/>.
```
rmarx, line 1231: I wonder if this is a good thing to say, as it has been shown that the defaults in RFC7540 (fair round robin) are typically the worst-case scenario for web page loading performance. 

Does it not make more sense to deprecate these defaults as well and instead recommend a FIFO/sequential scheme instead in the absence of any signals (or just refer to the extensible priorities draft for the better default options (which also uses FIFO))?

rmarx (commented):
I don't feel super-strongly about any of these, but I feel that given the problems with H2 prioritization, it's good to be as nuanced/concrete as possible.

draft-thomson-httpbis-http2bis.xml
martinthomson: size is just another aspect of content :)

draft-thomson-httpbis-http2bis.xml
martinthomson: How about s/widely/uniformly ?

draft-thomson-httpbis-http2bis.xml
martinthomson: Good call.  I was being conservative, but I think that we could easily make a stronger statement here.

draft-thomson-httpbis-http2bis.xml
martinthomson: I've tweaked this to say "if you receive priority signals from HEADERS or PRIORITY frames", which should be clear enough.

draft-thomson-httpbis-http2bis.xml
martinthomson: Done.  Thanks.

draft-thomson-httpbis-http2bis.xml
rmarx: Fair enough. It's mainly that size is a parameter typically available only at the server (and is often used in shortest-completion-time schedulers for e.g., multipath) and content (or at least expected content and its semantics/impact) is something typically mainly derived at the client (e.g., difference between render blocking, async or defer JS). 

Now, I don't think it's really necessary to make those nuances here, so I'm fine with dropping it :)

draft-ietf-httpbis-http2bis.xml
```diff
@@ -1242,24 +1242,24 @@
...
             alternative scheme, such as <xref target="I-D.ietf-httpbis-priority"/>.
           </t>
           <t>
             Though the priority signaling from RFC 7540 was not widely adopted, the information it
             provides can still be useful in the absence of better information.  Endpoints that
-            receive <xref target="HEADERS" format="none">HEADERS</xref> or <xref target="PRIORITY"
-            format="none">PRIORITY</xref> frames can benefit from applying any priority that is
-            present.  In particular, implementations that consume these signals would not benefit
-            from discarding these priority signals in the absence of alternatives.
+            receive priority signals <xref target="HEADERS" format="none">HEADERS</xref> or <xref
```
rmarx, line 1252: Should there be an "in" or "from" between  signals   and   HEADERS?

rmarx (approved):
These address all my points nicely, thanks!

martinthomson:
Thanks for the review Robin.  I think that this is much stronger as a result.

ianswett (approved):

LPardue (approved):

draft-ietf-httpbis-http2bis.xml
```diff
@@ -1198,231 +1194,77 @@
...
-        </t>
+        <name>Prioritization</name>
         <t>
-          The purpose of prioritization is to allow an endpoint to express how it would prefer its
-          peer to allocate resources when managing concurrent streams.  Most importantly, priority can
-          be used to select streams for transmitting frames when there is limited capacity for
-          sending.
+          In a multiplexed protocol like HTTP/2, prioritizing allocation of bandwidth and
+          computation resources to streams can be critical to attaining good performance.  A poor
+          prioritization scheme can result in HTTP/2 providing poor performance.  With no parallism
```
birenroy, line 1203: s/parallism/parallelism/

PR #797, opened 2020-12-22
Changes for TLS 1.3

### LINKED ISSUE #774: Integrate TLS 1.3 updates
A few small changes are included in [RFC 8740](https://tools.ietf.org/html/rfc8740), which could be integrated easily.

martinthomson:
A lot has happened to TLS since we did RFC 7540.  This tries to capture
that succinctly.

- It rolls in changes from RFC 8740.

- It mentions early data and RFC 8470.

- It updates references, citing TLS 1.3 rather than 1.2 for a generic
  TLS reference.

Closes #774.

### DIFF
--- a/draft-thomson-httpbis-http2bis.xml
+++ b/draft-thomson-httpbis-http2bis.xml
@@ -13,7 +13,7 @@
 <?rfc-ext allow-markup-in-artwork="yes" ?>
 <?rfc-ext include-index="no" ?>
 <!DOCTYPE rfc SYSTEM "rfc2629-xhtml.ent">
-<rfc xmlns:xi="http://www.w3.org/2001/XInclude" xmlns:x="http://purl.org/net/xml2rfc/ext" ipr="trust200902" category="std" docName="draft-thomson-httpbis-http2bis-latest" tocInclude="true" symRefs="true" sortRefs="true" version="3" submissionType="IETF" obsoletes="7540">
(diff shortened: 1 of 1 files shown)


draft-thomson-httpbis-http2bis.xml
```diff
@@ -3569,6 +3570,37 @@
             the application protocol and cipher suite are independently selected.
           </t>
         </section>
+        <section anchor="tls13">
```
Lukasa, line 3573: Minor editorial note: is it useful to use the `”tls13features”` anchor for consistency with 1.2 and so we don’t have two almost-identical anchors (“TLS13” and “tls13”) in the document?

draft-thomson-httpbis-http2bis.xml
```diff
@@ -3569,6 +3570,37 @@
...
+          </t>
+          <t>
+            <xref target="TLS13"/> defines other post-handshake messages, NewSessionTicket and
+            KeyUpdate, which can be used as they have no direct interaction with HTTP/2.  Unless the
+            use of a new type of TLS message depends on an interaction with the application-layer
+            protocol, that TLS message can be sent after the handshake completes.
+          </t>
+	  <t>
+	    TLS early data MAY be used to send requests, provided that the guidance in <xref
+	    target="RFC8470"/> is observed.  Clients send requests in early data assuming initial
```
Lukasa, line 3600: As this document obsoletes RFC 8470, is it worth importing that guidance into this document wholesale?

Lukasa (approved):

draft-thomson-httpbis-http2bis.xml
martinthomson: It's RFC 8740 that this obsoletes :)  Confusing, I know.  I have to double-check every time because I keep getting them the wrong way around.

Lukasa: That’s a huge mess.

PR #798, opened 2020-12-22
Condition SNI requirement more carefully

### LINKED ISSUE #792: The requirement to use SNI is not possible always
SNI contains a domain name.  Clients connecting to an host in order to make requests for a URL with an IP address authority don't include SNI.  We addressed this better in HTTP/3.

> If the server is identified by a DNS name, clients MUST send the Server Name Indication (SNI; [RFC6066]) TLS extension unless an alternative mechanism to indicate the target host is used.

-- https://quicwg.org/base-drafts/draft-ietf-quic-http.html#section-3.2-2

I believe that in practice the requirement is being ignored.  Aligning the spec with reality seems best.

martinthomson:
Closes #792.

### DIFF
--- a/draft-thomson-httpbis-http2bis.xml
+++ b/draft-thomson-httpbis-http2bis.xml
@@ -3482,8 +3482,9 @@
         </t>
         <t>
           The TLS implementation MUST support the <xref target="TLS-EXT">Server Name Indication
-          (SNI)</xref> extension to TLS. HTTP/2 clients MUST indicate the target domain name when
-          negotiating TLS.
+          (SNI)</xref> extension to TLS.  If the server is identified by a DNS name, clients MUST
+          send the server_name TLS extension unless an alternative mechanism to indicate the target
+          host is used.
         </t>
         <t>
           Deployments of HTTP/2 that negotiate TLS 1.3 or higher need only support and use the SNI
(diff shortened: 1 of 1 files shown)


Lukasa (approved):

draft-thomson-httpbis-http2bis.xml
```diff
@@ -3482,8 +3482,9 @@
         </t>
         <t>
           The TLS implementation MUST support the <xref target="TLS-EXT">Server Name Indication
-          (SNI)</xref> extension to TLS. HTTP/2 clients MUST indicate the target domain name when
-          negotiating TLS.
+          (SNI)</xref> extension to TLS.  If the server is identified by a DNS name, clients MUST
```
mnot, line 3485: Is "DNS name" the right terminology here? 3986 uses [reg-name](https://tools.ietf.org/html/rfc3986#section-3.2.2).

martinthomson: I think so.  `reg-name` is a broad umbrella that covers a bunch of things that might include other name resolution systems (NetBIOS, appletalk, /etc/hosts, namecoin, onion names).  And RFC 6066 uses "host name", which implies that it is also generic.  We could use "host name" instead, I guess.  That would be more generic, but it also implies a stronger requirement that might not be appropriate for those other uses.  I'm certain about DNS names, but I can't say for sure about the others.

mnot: I think your reasoning is sound. Is 'DNS name' formally defined somewhere? If so, we should probably reference.

martinthomson: I found something in RFC 8499 that seems right, but it means changing to "domain name".  I'll propose the same for QUIC, where this text comes from originally.

PR #800, opened 2020-12-22
Reference RFC 6973 for fingerprinting

### LINKED ISSUE #771: Update HTML5 reference
The current reference is in the WHATWG "infra standard": https://infra.spec.whatwg.org/#tracking-vector

To use this, we need to ask for the "tracking-vector" anchor to be made permanent.

martinthomson:
The thinking has evolved considerably since the original reference was
made.  This is the right way to refer to this concept.

Closes #771.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -3840,8 +3840,7 @@
         </t>
         <t>
           As far as these create observable differences in behavior, they could be used as a basis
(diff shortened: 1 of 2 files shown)


Lukasa (approved):

martinthomson:
So the re-request review button doesn't appear to work; @Lukasa, any objections to landing this with the new reference target?

Lukasa (approved):
None whatsoever.

PR #806, opened 2021-01-21
Remove use of "payload" for message bodies.

### LINKED ISSUE #803: "Payload" is too heavily overloaded
See https://github.com/httpwg/http-core/issues/654

Lukasa:
This patch also changes our text to be consistent in using "frame
payload" as a two-word phrase.

Note that this patch now sometimes leaves us with the text "frame payload of an X frame". Shout if that seems too clumsy and I can remove the last word "frame", though I think it has the advantage of being extremely clear.

Resolves #803.

### DIFF
--- a/draft-thomson-httpbis-http2bis.xml
+++ b/draft-thomson-httpbis-http2bis.xml
@@ -2705,7 +2705,7 @@
               frames) containing the message headers (see <xref target="RFC7230" section="3.2"/>),
             </li>
           <li>
-              zero or more <xref target="DATA" format="none">DATA</xref> frames containing the content (see <xref target="draft-ietf-httpbis-semantics-14" section="6.4"/>), and
+              zero or more <xref target="DATA" format="none">DATA</xref> frames containing the message content (see <xref target="draft-ietf-httpbis-semantics-14" section="6.4"/>), and
             </li>
           <li>
               optionally, one <xref target="HEADERS" format="none">HEADERS</xref> frame, followed by zero or more
--- a/draft-thomson-httpbis-http2bis.xml
+++ b/draft-thomson-httpbis-http2bis.xml
@@ -4781,48 +4781,18 @@
                     surname="Fielding"
                     role="editor">
               <organization>Adobe</organization>
-              <address>
-                <postal>
-                  <street>345 Park Ave</street>
-                  <city>San Jose</city>
-                  <region>CA</region>
-                  <code>95110</code>
-                  <country>United States of America</country>
-                </postal>
-                <email>fielding@gbiv.com</email>
-                <uri>https://roy.gbiv.com/</uri>
-              </address>
             </author>
             <author fullname="Mark Nottingham"
                     initials="M."
                     surname="Nottingham"
                     role="editor">
               <organization>Fastly</organization>
-              <address>
-                <postal>
-                  <city>Prahran</city>
-                  <region>VIC</region>
-                  <country>Australia</country>
-                </postal>
-                <email>mnot@mnot.net</email>
-                <uri>https://www.mnot.net/</uri>
-              </address>
             </author>
             <author fullname="Julian Reschke"
                     initials="J."
                     surname="Reschke"
                     role="editor">
               <organization abbrev="greenbytes">greenbytes GmbH</organization>
-              <address>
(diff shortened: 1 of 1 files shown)


draft-thomson-httpbis-http2bis.xml
```diff
@@ -2705,7 +2705,7 @@
               frames) containing the message headers (see <xref target="RFC7230" section="3.2"/>),
             </li>
           <li>
-              zero or more <xref target="DATA" format="none">DATA</xref> frames containing the payload body (see <xref target="RFC7230" section="3.3"/>), and
+              zero or more <xref target="DATA" format="none">DATA</xref> frames containing the content (see <xref target="draft-ietf-httpbis-semantics-14" section="6.4"/>), and
```
martinthomson, line 2708: Optional:
```suggestion
              zero or more <xref target="DATA" format="none">DATA</xref> frames containing the message content (see <xref target="draft-ietf-httpbis-semantics-14" section="6.4"/>), and
```

draft-thomson-httpbis-http2bis.xml
```diff
@@ -3163,9 +3163,9 @@
         </t>
         <t>
           Promised requests MUST be cacheable (see <xref target="RFC7231" section="4.2.3"/>), MUST be safe
-          (see <xref target="RFC7231" section="4.2.1"/>), and MUST NOT include a request body. Clients that receive a
+          (see <xref target="RFC7231" section="4.2.1"/>), and MUST NOT include any content. Clients that receive a
```
martinthomson, line 3166: Separately, I'm wondering if this "MUST NOT" is better as "cannot" to match the "is not possible" below.

draft-thomson-httpbis-http2bis.xml
```diff
@@ -4773,6 +4773,61 @@
...
+                <postal>
+                  <street>345 Park Ave</street>
+                  <city>San Jose</city>
+                  <region>CA</region>
+                  <code>95110</code>
+                  <country>United States of America</country>
+                </postal>
+                <email>fielding@gbiv.com</email>
+                <uri>https://roy.gbiv.com/</uri>
+              </address>
```
martinthomson, line 4794:
```suggestion
```

draft-thomson-httpbis-http2bis.xml
```diff
@@ -4773,6 +4773,61 @@
...
+              <organization>Fastly</organization>
+              <address>
+                <postal>
+                  <city>Prahran</city>
+                  <region>VIC</region>
+                  <country>Australia</country>
+                </postal>
+                <email>mnot@mnot.net</email>
+                <uri>https://www.mnot.net/</uri>
+              </address>
```
martinthomson, line 4809:
```suggestion
```

draft-thomson-httpbis-http2bis.xml
```diff
@@ -4773,6 +4773,61 @@
...
+              <address>
+                <postal>
+                  <street>Hafenweg 16</street>
+                  <city>Münster</city>
+                  <code>48155</code>
+                  <country>Germany</country>
+                </postal>
+                <email>julian.reschke@greenbytes.de</email>
+                <uri>https://greenbytes.de/tech/webdav/</uri>
+              </address>
```
martinthomson, line 4825:
```suggestion
```

martinthomson (approved):

draft-thomson-httpbis-http2bis.xml
Lukasa: Hmm, which "is not possible"? I don't see an is-not-possible anywhere near here.

martinthomson: > It is not possible to push a response to a request that includes message content.

In the next bit of diff.  Not as close in the file though.

Lukasa: Ah, I see. My mistake for looking through the file. :wink: But I see what you're driving at.

I think we need at least one normative statement that promised requests must not contain content. Otherwise, saying "promised requests cannot include any content" invites the question "well, why not?". Having at least one normative statement means the answer is "because we said so" and that's a sufficiently good answer for spec purposes. So I'm inclined to leave it.

PR #807, opened 2021-01-21
Defer to the core spec for connection-specific headers

### LINKED ISSUE #804: Connection headers and proxies
We have generation rules
> An endpoint MUST NOT generate an HTTP/2 message containing connection-specific header fields; any message containing connection-specific header fields MUST be treated as malformed (Section 8.1.2.6).

And we have forwarding rules:
> This means that an intermediary transforming an HTTP/1.x message to HTTP/2 will need to remove any header fields nominated by the Connection header field, along with the Connection header field itself.  Such intermediaries SHOULD also remove other connection-specific header fields, such as Keep-Alive, Proxy-Connection, Transfer-Encoding, and Upgrade, even if they are not nominated by the Connection header field.

In https://github.com/quicwg/base-drafts/pull/4747 for HTTP/3, we are discussing a requirement to remove Transfer-Encoding.  But the "will need to" for the proxy is what seems most strange.  Should this be a MUST instead?

Lukasa:
Resolves #804.

### DIFF
--- a/draft-thomson-httpbis-http2bis.xml
+++ b/draft-thomson-httpbis-http2bis.xml
@@ -2820,11 +2820,9 @@
               request; when it is, it MUST NOT contain any value other than "trailers".
             </t>
             <t>
-              This means that an intermediary transforming an HTTP/1.x message to HTTP/2 will need
-              to remove any header fields nominated by the Connection header field, along with the
-              Connection header field itself. Such intermediaries SHOULD also remove other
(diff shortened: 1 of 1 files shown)


Lukasa:
Note: this patch depends on #806 for the xref. I don't want to add it here as well because then they won't merge cleanly.

martinthomson:
WFM.  We can keep this open until https://github.com/quicwg/base-drafts/pull/4747 lands so that the language stays coherent between the two specs.

PR #808, opened 2021-01-21
Remove definition of 421 and point to core.

### LINKED ISSUE #802: Defer to core HTTP for 421 status code
https://github.com/httpwg/http-core/pull/624 added the definition of 421 (Misdirected Request) to the core HTTP specification.  We don't need to replicate that definition any more, it is only necessary to keep pointers.

Lukasa:
Resolves #802.

Depends on #806 for the xref, won't build cleanly without it. However, I don't want to add the xref here as it would prevent a clean merge.


martinthomson (approved):
Let's get #806 in first :)

PR #811, opened 2021-02-02
First pass on references to updated documents

martinthomson:
This is fairly cursory at this stage.  The goal is to get the document
functional first.  Next step is a full read-through for terminology.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -620,12 +620,13 @@
         </t>
       </section>
       <section anchor="FieldBlock">
-        <name>Header Compression and Decompression</name>
+        <name>Field Section Compression and Decompression</name>
         <t>
-          The term "header compression", or its inverse of "header decompression", is used to refer
-          to the process of compressing, or decompressing, a set of field lines to form a field
-          block.  The process HTTP/2 uses for header compression and decompression is defined in
-          <xref target="COMPRESSION"/>.
+          Field section compression is the process of compressing a set of field lines to form a
+          field block.  Field section decompression is the process of decoding a field block into a
+          set of field lines.  Details of HTTP/2 field section compression and decompression is
+          defined in <xref target="COMPRESSION"/>, which, for historical reasons, refers to these
+          processes as header compression and decompression.
         </t>
         <t>
           Field blocks carry the compressed bytes of a field section, with header sections
@@ -1480,7 +1481,7 @@
             The peer that sends the <xref target="RST_STREAM" format="none">RST_STREAM</xref> frame MUST be prepared to receive any
             frames that were sent or enqueued for sending by the remote peer.  These frames can be
             ignored, except where they modify connection state (such as the state maintained for
-            <xref target="FieldBlock">header compression</xref> or flow control).
+            <xref target="FieldBlock">field section compression</xref> or flow control).
           </t>
           <t>
             Normally, an endpoint SHOULD NOT send more than one <xref target="RST_STREAM" format="none">RST_STREAM</xref> frame for
@@ -2183,7 +2184,7 @@
         </t>
         <t>
           A PUSH_PROMISE frame modifies the connection state in two ways.  First, the inclusion of a <xref target="FieldBlock">field block</xref> potentially modifies the state maintained for
-          header compression.  Second, PUSH_PROMISE also reserves a stream for later use, causing the
+          field section compression.  Second, PUSH_PROMISE also reserves a stream for later use, causing the
           promised stream to enter the "reserved" state.  A sender MUST NOT send a PUSH_PROMISE on a
           stream unless that stream is either "open" or "half-closed (remote)"; the sender MUST
           ensure that the promised stream is a valid choice for a <xref target="StreamIdentifiers">new stream identifier</xref> (that is, the promised stream MUST
@@ -2375,10 +2376,10 @@
           receiver with identifiers higher than the identified last stream.  However, any frames
           that alter connection state cannot be completely ignored.  For instance,
           <xref target="HEADERS" format="none">HEADERS</xref>, <xref target="PUSH_PROMISE" format="none">PUSH_PROMISE</xref>, and <xref target="CONTINUATION" format="none">CONTINUATION</xref> frames
-          MUST be minimally processed to ensure the state maintained for header compression is
+          MUST be minimally processed to ensure the state maintained for field section compression is
           consistent (see <xref target="FieldBlock"/>); similarly, DATA frames MUST be counted
           toward the connection flow-control window.  Failure to process these frames can cause flow
-          control or header compression state to become unsynchronized.
+          control or field section compression state to become unsynchronized.
         </t>
         <t>
           The GOAWAY frame also contains a 32-bit <xref target="ErrorCodes">error code</xref> that
@@ -2681,7 +2682,8 @@
           </dd>
         <dt>COMPRESSION_ERROR (0x9):</dt>
         <dd anchor="COMPRESSION_ERROR">
-            The endpoint is unable to maintain the header compression context for the connection.
+            The endpoint is unable to maintain the field section compression context for the
+            connection.
           </dd>
         <dt>CONNECT_ERROR (0xa):</dt>
         <dd anchor="CONNECT_ERROR">
@@ -3709,7 +3711,7 @@
         <name>Denial-of-Service Considerations</name>
         <t>
           An HTTP/2 connection can demand a greater commitment of resources to operate than an
-          HTTP/1.1 connection.  The use of header compression and flow control depend on a
+          HTTP/1.1 connection.  The use of field section compression and flow control depend on a
           commitment of resources for storing a greater amount of state.  Settings for these
           features ensure that memory commitments for these features are strictly bounded.
         </t>
@@ -3737,7 +3739,7 @@
           end of a stream.
         </t>
         <t>
-          Header compression also offers some opportunities to waste processing resources; see
+          Field section compression also offers some opportunities to waste processing resources; see
           <xref target="COMPRESSION" section="7"/> for more details on potential abuses.
         </t>
         <t>
@@ -3747,9 +3749,9 @@
           known to clients and could be exceeded without being an obvious protocol violation.
         </t>
         <t>
-          All these features -- i.e., <xref target="SETTINGS" format="none">SETTINGS</xref> changes, small frames, header
-          compression -- have legitimate uses.  These features become a burden only when they are
-          used unnecessarily or to excess.
+          All these features -- i.e., <xref target="SETTINGS" format="none">SETTINGS</xref> changes,
+          small frames, field section compression -- have legitimate uses.  These features become a
+          burden only when they are used unnecessarily or to excess.
         </t>
         <t>
           An endpoint that doesn't monitor this behavior exposes itself to a risk of denial-of-service
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -141,7 +141,7 @@
       </t>
       <t>
         Because HTTP header fields used in a connection can contain large amounts of redundant
-        data, frames that contain them are <xref target="HeaderBlock">compressed</xref>. This has
+        data, frames that contain them are <xref target="FieldBlock">compressed</xref>. This has
         especially advantageous impact upon request sizes in the common case, allowing many
         requests to be compressed into one packet.
       </t>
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2bis.xml
```diff
@@ -71,7 +71,7 @@
       <name>Introduction</name>
       <t>
         The Hypertext Transfer Protocol (HTTP) is a wildly successful protocol. However, the way
-        HTTP/1.1 uses the underlying transport (<xref target="RFC7230" section="6"/>) has several
+        HTTP/1.1 uses the underlying transport (<xref target="HTTP11"/>) has several
```
Lukasa, line 74: We really don't have a section reference here?

draft-ietf-httpbis-http2bis.xml
```diff
@@ -2688,19 +2688,11 @@
...
-        that, from the application perspective, the features of the protocol are largely
-        unchanged. To achieve this, all request and response semantics are preserved, although the
-        syntax of conveying those semantics has changed.
-      </t>
-      <t>
-        Thus, the specification and requirements of HTTP/1.1 Semantics and Content <xref target="RFC7231"/>, Conditional Requests <xref target="RFC7232"/>, Range Requests <xref target="RFC7233"/>, Caching <xref target="RFC7234"/>, and Authentication <xref target="RFC7235"/> are applicable to HTTP/2. Selected portions of HTTP/1.1 Message Syntax
-        and Routing <xref target="RFC7230"/>, such as the HTTP and HTTPS URI schemes, are also
-        applicable in HTTP/2, but the expression of those semantics for this protocol are defined
-        in the sections below.
+        HTTP/2 defines a framing of the abstract HTTP message abstraction
```
Lukasa, line 2691: Editorial nit: "abstract HTTP message abstraction". Aren't abstractions tautologically abstract? I think I'd propose deleting the first instance of the word "abstract".

```suggestion
        HTTP/2 defines a framing of the HTTP message abstraction
```

draft-ietf-httpbis-http2bis.xml
```diff
@@ -2710,23 +2702,32 @@
...
-              or more <xref target="CONTINUATION" format="none">CONTINUATION</xref> frames) containing the message headers of
-              informational (1xx) HTTP responses (see <xref target="RFC7230" section="3.2"/> and
-              <xref target="RFC7231" section="6.2"/>),
-            </li>
-          <li>
-              one <xref target="HEADERS" format="none">HEADERS</xref> frame (followed by zero or more <xref target="CONTINUATION" format="none">CONTINUATION</xref>
-              frames) containing the message headers (see <xref target="RFC7230" section="3.2"/>),
+              one <xref target="HEADERS" format="none">HEADERS</xref> frame (followed by zero or
+              more <xref target="CONTINUATION" format="none">CONTINUATION</xref> frames) containing
+              the header section (see <xref target="HTTP" section="5"/>),
```
Lukasa, line 2707: I'm not sure about this reference. Section 5 is "Fields", but this doesn't define "header section" (though perplexingly does _use_ the term). "Header section" is defined in § 6.3. I think we should be referring to that instead. This also lines up nicely with the other references in this list, which point to § 6.4 and § 6.5.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -2710,23 +2702,32 @@
...
+              section="6.5"/>).
             </li>
         </ol>
+        <t>
+          For a response only, a server MAY send any number of interim responses before the <xref
+          target="HEADERS" format="none">HEADERS</xref> frame containing a final response.  An
+          interim response consists of a <xref target="HEADERS" format="none">HEADERS</xref> frames
+          (which might be followed by zero or more <xref target="CONTINUATION"
+          format="none">CONTINUATION</xref> frames) containing the control data and header section
+          of an interim (1xx) HTTP responses (see <xref target="HTTP" section="15"/>).  A <xref
```
Lukasa, line 2726: Should this reference be more specific, to § 15.2? § 15 is "Status codes", § 15.2 is "Informational 1xx".

draft-ietf-httpbis-http2bis.xml
```diff
@@ -2976,7 +2978,7 @@
               A request or response that includes message content can include a <tt>content-length</tt> header field.  A request or response is also
               malformed if the value of a <tt>content-length</tt> header field
               does not equal the sum of the <xref target="DATA" format="none">DATA</xref> frame payload lengths that form the
-              content.  A response that is defined to have no content, as described in <xref target="RFC7230"/>, can have a non-zero
+              content.  A response that is defined to have no content, as described in <xref target="HTTP"/>, can have a non-zero
```
Lukasa, line 2981: Can we make this more specific? I think we could point at § 6.4 ("Content").

draft-ietf-httpbis-http2bis.xml
```diff
@@ -3181,17 +3183,18 @@
...
+          target="StreamErrorHandler">stream error</xref> of type <xref target="PROTOCOL_ERROR"
+          format="none">PROTOCOL_ERROR</xref>.  Note this could result in the promised stream being
+          reset if the client does not recognize a newly defined method as being safe.
         </t>
         <t>
-          Pushed responses that are cacheable (see <xref target="RFC7234" section="3"/>) can be stored by the client, if it implements an HTTP
+          Pushed responses that are cacheable (see <xref target="CACHE" section="3"/>) can be stored by the client, if it implements an HTTP
           cache.  Pushed responses are considered successfully validated on the origin server (e.g.,
-          if the "no-cache" cache response directive is present; see <xref target="RFC7234" section="5.2.2"/>) while the stream identified by the
+          if the "no-cache" cache response directive is present; see <xref target="CACHE" section="5.2.2.4"/>) while the stream identified by the
```
Lukasa, line 3197: no-cache is § 5.2.2.3, not § 5.2.2.4.

```suggestion
          if the "no-cache" cache response directive is present; see <xref target="CACHE" section="5.2.2.3"/>) while the stream identified by the
```

Lukasa (approved):
Generally looks good, some notes on specific references in the diff.

draft-ietf-httpbis-http2bis.xml
martinthomson: It's a very broad reference, it doesn't need to be precise.

martinthomson (commented):
Note that one non-obvious thing here is that HTTP/1.1 is a normative reference here.  I didn't think that it needed to be initially, but then I found that it was unavoidable because we talk a LOT about translating too and from 1.1.  I don't think that we can avoid that unfortunately.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -2021,10 +2022,10 @@
             <dt anchor="SETTINGS_MAX_HEADER_LIST_SIZE">SETTINGS_MAX_HEADER_LIST_SIZE (0x6):</dt>
             <dd>
               <t>
-                  This advisory setting informs a peer of the maximum size of header list that the
+                  This advisory setting informs a peer of the maximum size of header section that the
```
afrind, line 2025: Does this setting also apply to trailer sections?

draft-ietf-httpbis-http2bis.xml
```diff
@@ -2159,7 +2160,7 @@
           the PUSH_PROMISE.
         </t>
         <t>
-          A PUSH_PROMISE frame modifies the connection state in two ways.  First, the inclusion of a <xref target="HeaderBlock">header block</xref> potentially modifies the state maintained for
+          A PUSH_PROMISE frame modifies the connection state in two ways.  First, the inclusion of a <xref target="HeaderBlock">field block</xref> potentially modifies the state maintained for
           header compression.  Second, PUSH_PROMISE also reserves a stream for later use, causing the
```
afrind, line 2164: Is 'header compression' still what you want here?

draft-ietf-httpbis-http2bis.xml
```diff
@@ -3086,15 +3091,15 @@
...
           </t>
           <t>
-            Trailing header fields are sent as a header block after both the request or response
-            header block and all the <xref target="DATA" format="none">DATA</xref> frames have been sent.  The
-            <xref target="HEADERS" format="none">HEADERS</xref> frame starting the trailers header block has the END_STREAM flag
-            set.
+            A trailer section is sent as a field block after both the request or response
+            field block and all the <xref target="DATA" format="none">DATA</xref> frames have been sent.  The
+            <xref target="HEADERS" format="none">HEADERS</xref> frame starting the field block that comprises
+            the trailer section has the END_STREAM flag set.
```
afrind, line 3097: Is this a MUST and/or is the message malformed otherwise?

draft-ietf-httpbis-http2bis.xml
```diff
@@ -3732,29 +3736,29 @@
...
-            commit a large amount of available memory for header fields.
+            correctness, mean that an endpoint might need to buffer the entire field block.  Since
+            there is no hard limit to the size of a field block, some endpoints could be forced to
+            commit a large amount of available memory for field blocks.
           </t>
           <t>
             An endpoint can use the <xref target="SETTINGS_MAX_HEADER_LIST_SIZE" format="none">SETTINGS_MAX_HEADER_LIST_SIZE</xref> to advise peers of
-            limits that might apply on the size of header blocks.  This setting is only advisory, so
-            endpoints MAY choose to send header blocks that exceed this limit and risk having the
+            limits that might apply on the size of field blocks.  This setting is only advisory, so
```
afrind, line 3751: The setting applies to the size of a decompressed header list, which implicitly limits the size of the blocks (it is compression, after all).  Does that need to be called out more explicitly?

draft-ietf-httpbis-http2bis.xml
```diff
@@ -619,28 +620,28 @@
         </t>
       </section>
       <section anchor="HeaderBlock">
-        <name>Header Compression and Decompression</name>
+        <name>Field Section Compression and Decompression</name>
         <t>
-          Just as in HTTP/1, a header field in HTTP/2 is a name with one or more associated values.
-          Header fields are used within HTTP request and response messages as well as in server push operations
+          A field in HTTP is a name with one or more associated values.
+          Field lines carry each value and are used within HTTP/2 request and response messages as well as in server push operations
```
afrind, line 626: HPACK compresses both fields and control data and applies to "field lines", but control data is not a field?  This might need more explanation.

Lukasa: Yeah I think this specific change doesn't read right to me. Is "field block" really right? That feels very strange.

draft-ietf-httpbis-http2bis.xml
Lukasa: I thought we called this the "field section" above.

draft-ietf-httpbis-http2bis.xml
martinthomson: I think that it is still OK to call it header compression, especially as we aren't revising RFC 7541.  It is a little janky and it probably needs better explanation, but I don't see a better alternative.

draft-ietf-httpbis-http2bis.xml
martinthomson: There are MUSTs in other places.

draft-ietf-httpbis-http2bis.xml
martinthomson: I've given this another go, with a more comprehensive rewrite of the first few paragraphs of this section.  Let me know if that reads better.

Lukasa: I do think it's better, yeah.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -619,18 +619,40 @@
...
-      <section anchor="HeaderBlock">
-        <name>Field Section Compression and Decompression</name>
+      <section anchor="FieldBlock">
+        <name>Header Compression and Decompression</name>
         <t>
-          A field in HTTP is a name with one or more associated values.
-          Field lines carry each value and are used within HTTP/2 request and response messages as well as in server push operations
-          (see <xref target="PushResources"/>).
+          The term "header compression", or its inverse of "header decompression", is used to refer
+          to the process of compressing, or decompressing, a set of field lines to form a field
```
afrind, line 626: 'or decompressing' reads weirdly to me, since that doesn't apply to the action described: field lines -> field block.  Might be better to just say compression transforms field lines to field blocks and decompression transforms field blocks into field lines.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -619,18 +619,40 @@
...
+          block.  The process HTTP/2 uses for header compression and decompression is defined in
+          <xref target="COMPRESSION"/>.
         </t>
         <t>
-          A field section is a collection of zero or more field lines.  When transmitted over a
-          connection, a field section is serialized into a field block using <xref target="COMPRESSION">HTTP header compression</xref> (previous versions of this specification used the term "header block").  The serialized field block is then
-          divided into one or more octet sequences, called field block fragments, and transmitted
-          within the frame payload of <xref target="HEADERS">HEADERS</xref>, <xref target="PUSH_PROMISE">PUSH_PROMISE</xref>, or <xref target="CONTINUATION">CONTINUATION</xref> frames.
+          Field blocks carry the compressed bytes of a field section, with header sections
```
afrind, line 631: would 'header field sections' read better?

draft-ietf-httpbis-http2bis.xml
```diff
@@ -619,18 +619,40 @@
           which, if blocked by the transmission of a large frame, could affect performance.
         </t>
       </section>
-      <section anchor="HeaderBlock">
-        <name>Field Section Compression and Decompression</name>
+      <section anchor="FieldBlock">
+        <name>Header Compression and Decompression</name>
```
afrind, line 623: I feel like "Field Section Compression and Decompression" would be more consistent.

draft-ietf-httpbis-http2bis.xml
afrind: I think we can call it field section compression everywhere in this document, and explain in the terms section how 7541 Header Compression applies?

martinthomson:
"field section compression" it is.

martinthomson:
This might need a double-check later, but it's too big to leave sitting.

Issue #759, opened 2017-01-14
Stream State and PRIORITY


Scottmitch:
What is the expected result in terms of stream state and any potential errors that should be generated from the following scenario:

```
1. Client [PRIORITY, id: 5] -> Server
2. Client [HEADERS, id: 3] -> Server
```

## Option 1
The server accepts both frames. Stream `3` is OPEN and stream `5` remains in IDLE.

This behavior is verified in the h2spec verification tool via https://github.com/summerwind/h2spec/pull/67. Also see additional discussion there.

## Option 2
Should the use of stream ID `5` cause stream ID `3` (and all lower stream IDs) to be implicitly closed according to the section-5.1.1 (see [1])? This would mean after the server receives the frame from step (1) it thinks that stream ID `3` has been closed, and when the server receives the frame from step (2) it will respond with a connection error of type PROTOCOL_ERROR (see [2])?

[1] https://tools.ietf.org/html/rfc7540#section-5.1.1
> The first use of a new stream identifier implicitly closes all
   streams in the "idle" state that might have been initiated by that
   peer with a lower-valued stream identifier.

[2] https://tools.ietf.org/html/rfc7540#section-5.1.1
> An endpoint that
   receives an unexpected stream identifier MUST respond with a
   connection error (Section 5.4.1) of type PROTOCOL_ERROR.

Scottmitch:
This is being discussed on the httpbis mailing list https://lists.w3.org/Archives/Public/ietf-http-wg/2017JanMar/0046.html#start46

mnot:
Leaving it open for bookkeeping.

martinthomson:
Unsurprisingly, @Lukasa had what I think is [the definitive answer](https://lists.w3.org/Archives/Public/ietf-http-wg/2017JanMar/0049.html) from the thread in support of option 1 here.  We should try to clarify this further by making it clear that PRIORITY doesn't affect stream state (including for lower-numbered streams).

As this is now just editorial clarifications, I think that taking this as editorial is justified.  We can revisit if the changes start to look too scary.

Lukasa:
I believe we addressed this in #813.

martinthomson:
Agreed.

PR #813, opened 2021-02-03
Clarify that PRIORITY does not close streams.

Lukasa:
This is a first stab at #759. I've gone for a minimal change here, simply clarifying that only HEADERS and PUSH_PROMISE frames have the effect of implicitly closing streams.

While I was writing this I did have a question that may need to be raised on-list: do HEADERS and PUSH_PROMISE only implicitly close streams if they are _accepted_? That is, if a client sends a HEADERS frame that triggers a stream error, does that HEADERS frame still implicitly close all streams with a lower stream ID that are in the "idle" state? This may also be worth clarifying.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -1032,11 +1032,12 @@
             type <xref target="PROTOCOL_ERROR" format="none">PROTOCOL_ERROR</xref>.
           </t>
           <t>
-            The first use of a new stream identifier implicitly closes all streams in the "idle"
-            state that might have been initiated by that peer with a lower-valued stream identifier.
-            For example, if a client sends a <xref target="HEADERS" format="none">HEADERS</xref> frame on stream 7 without ever
-            sending a frame on stream 5, then stream 5 transitions to the "closed" state when the
-            first frame for stream 7 is sent or received.
+            A <xref target="HEADERS" format="none">HEADERS</xref> frame will transition the client-initiated stream identified
+            by the stream identifier in frame header from "idle" to "open". A <xref target="PUSH_PROMISE" format="none">PUSH_PROMISE</xref>
+            frame will transition the server-initiated stream identified by the "Promised Stream ID" field in the frame payload. When
+            a stream transitions out of the "idle" state, all streams that might have been initiated by that peer with a lower-valued
+            stream identifier are implicitly transitioned to "closed". That is, an endpoint may skip a stream identifier, with the
+            effect being that the skipped stream is immediately closed.
           </t>
           <t>
             Stream identifiers cannot be reused.  Long-lived connections can result in an endpoint
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -1033,8 +1033,8 @@
           </t>
           <t>
             A <xref target="HEADERS" format="none">HEADERS</xref> frame will transition the client-initiated stream identified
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2bis.xml
```diff
@@ -1032,7 +1032,8 @@
             type <xref target="PROTOCOL_ERROR" format="none">PROTOCOL_ERROR</xref>.
           </t>
           <t>
-            The first use of a new stream identifier implicitly closes all streams in the "idle"
+            The first use of a new stream identifier in a <xref target="HEADERS" format="none">HEADERS</xref>
+            or <xref target="PUSH_PROMISE" format="none">PUSH_PROMISE</xref> frame implicitly closes all streams in the "idle"
             state that might have been initiated by that peer with a lower-valued stream identifier.
```
martinthomson, line 1037: I've a little trouble with the ambiguity regarding stream identifiers in PUSH_PROMISE.  I'm not clear on this, but it might be better to say:

> Each type of stream transitions out of the "idle" state for one reason:
> * A client-initiated stream becomes "open" when a HEADERS frame is sent on the stream.
> * A server-initiated stream becomes "reserved" when a PUSH_PROMISE (sent on a request stream) promises a response on the stream.
> When a stream transitions out of the "idle" state, all streams in that direction with lower-numbered stream identifiers that are also "idle" are closed.  That is, an endpoint can skip a stream identifier, with the effect being that the skipped stream is immediately closed.

Too large a change?

Lukasa: No, I think that's a good change. I may try to call out the specific stream identifier in the PUSH_PROMISE frame payload that is relevant here.

Lukasa: Hmm, I abandoned that change a bit because it didn't read well with the paragraph above, so I've tried something else. Let me know how you feel about it.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -1032,11 +1032,12 @@
             type <xref target="PROTOCOL_ERROR" format="none">PROTOCOL_ERROR</xref>.
           </t>
           <t>
-            The first use of a new stream identifier implicitly closes all streams in the "idle"
-            state that might have been initiated by that peer with a lower-valued stream identifier.
-            For example, if a client sends a <xref target="HEADERS" format="none">HEADERS</xref> frame on stream 7 without ever
-            sending a frame on stream 5, then stream 5 transitions to the "closed" state when the
-            first frame for stream 7 is sent or received.
+            A <xref target="HEADERS" format="none">HEADERS</xref> frame will transition the client-initiated stream identified
+            by the stream identifier in frame header from "idle" to "open". A <xref target="PUSH_PROMISE" format="none">PUSH_PROMISE</xref>
```
martinthomson, line 1036:
```suggestion
            by the stream identifier in the frame header from "idle" to "open". A <xref target="PUSH_PROMISE" format="none">PUSH_PROMISE</xref>
```

draft-ietf-httpbis-http2bis.xml
```diff
@@ -1032,11 +1032,12 @@
...
+            frame will transition the server-initiated stream identified by the "Promised Stream ID" field in the frame payload. When
```
martinthomson, line 1037:
```suggestion
            frame will transition the server-initiated stream identified by the "Promised Stream ID" field in the frame payload from "idle" to "reserved". When
```

?

martinthomson (approved):
Like it.  Take the first, maybe take the second, and we can merge this one.

PR #814, opened 2021-02-03
Reduce the width of the upgrade diagram.

### LINKED ISSUE #810: A wide diagram might be made narrower
.../draft-ietf-httpbis-http2bis.xml(311): Warning: Artwork too wide, reducing indentation from 3 to 1

Might be worth trying to fix.

Lukasa:
Fixes #810.


martinthomson:
simple :)

PR #816, opened 2021-02-05
fix IETF working group name

reschke:
Let's try this again.


Lukasa (approved):
I'm as happy with this now as I was when I first reviewed it.

martinthomson:
OK, this is weird.  I need to re-check my remotes.

martinthomson:
apparently we were OK, but the working group name change is enough; changed the title and merged

PR #819, opened 2021-02-12
Clarify error handling when multiple errors are encountered.

### LINKED ISSUE #770: Correct handling of HTTP/2 frames with multiple errors
It's not clear to me from RFC 7540 how an endpoint should behave upon receiving a frame that is erroneous in more than one way.

For example, suppose a server receives the following frame when it has stream id 1 in half-closed (remote) state:

```
{
  length = 4
  type = HEADERS
  flags = PRIORITY
  stream id = 1
  E = 0
  stream dependency = 1
}
```

- According to §4.2, the frame "MUST be treated as a connection error" and the server "MUST send an error code of FRAME_SIZE_ERROR" (because the length of 4 "is too small to contain mandatory frame data", namely the weight and header block fragment fields);
- According to §5.1, the server "MUST respond with a stream error (Section 5.4.2) of type STREAM_CLOSED" (because a frame "other than WINDOW_UPDATE, PRIORITY, or RST_STREAM" was received for a stream that is in "half-closed (remote)" state); and
- According to §5.3.1, the server "MUST treat this as a stream error (Section 5.4.2) of type PROTOCOL_ERROR" (because "a stream cannot depend on itself").

That is to say, according to the RFC, the server MUST (at very least) respond to this frame with one GOAWAY frame and two RST_STREAM frames.  However, sending all these frames would violate other requirements:

- According to §5.4.1, "after sending the GOAWAY frame for an error condition, the endpoint MUST close the TCP connection".  This obviously prevents other frames from being sent.
- According to §5.1, sending a RST_STREAM transitions the stream to closed state, in which "an endpoint MUST NOT send frames other than PRIORITY".  This prevents any further RST_STREAM frames from being sent for the same stream.

In this particular case, it seems reasonable to me that the frame size error should halt further processing of the frame such that neither of the stream errors are even detected: after all, an incorrect frame size may indicate corruption (in which case the rest of the frame's data must also be suspect).  But this is neither clear from the spec nor might other cases be so clear cut.

I see three possibilities, with a reasonably strong leaning toward the first:

1. Clarify that further processing of frames SHOULD (or MUST?) be halted upon encountering an error (I suspect this is what most implementations currently do, irrespective that the RFC says they "MUST" behave otherwise?).  But then the order of processing becomes significant: is it right that this should be left as an implementation detail, or should it be specified?  After all, the above frame might trigger a STREAM_CLOSED stream error from one server and a FRAME_SIZE_ERROR connection error from another, with very different consequences: how might these differences impact intermediaries, caching and retries?  Better I think for the order to be specified.

2. Define a precedence to errors, and mandate that that with highest precedence be sent (for example, connection errors might have higher precedence than stream errors).  But then endpoints must continue processing frames they know to be erroneous simply to discover whether any higher precedence errors also exist (at least until an error of the highest possible precedence is encountered).

3. Permit (require?) all errors arising from a single frame to be transmitted irrespective of any other errors the frame may also have triggered.  Again, endpoints would then continue processing frames that they know to be erroneous—but the peer would then be fully informed of all problems in order that it is better able to recover.

Lukasa:
Resolves #770.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -1428,7 +1428,7 @@
         </t>
         <t>
           If an endpoint detects multiple different errors,
-          it MAY choose to report any one of those errors. If a frame would cause a connection error, that error MUST be reported.
+          it MAY choose to report any one of those errors. If a frame causes a connection error, that error MUST be reported.
         </t>
         <section anchor="ConnectionErrorHandler">
           <name>Connection Error Handling</name>
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -1415,6 +1415,21 @@
         <t>
           A list of error codes is included in <xref target="ErrorCodes"/>.
         </t>
+        <t>
+          It is possible that an endpoint will encounter frames that would cause multiple errors. Implementations MAY discover
+          multiple errors during processing, but they SHOULD report at most one stream and one connection error as a result.
+        </t>
+        <t>
+          The first stream error reported for a given stream prevents any other errors on that stream from being reported.
+          In comparison, the protocol permits multiple <xref target="GOAWAY" format="none">GOAWAY</xref> frames, though an
+          endpoint SHOULD report just one type of connection error unless an error is encountered during graceful shutdown.
+          If this occurs, an endpoint MAY send an additional GOAWAY frame with the new error code, in addition to any prior
+          GOAWAY that contained <xref target="NO_ERROR" format="none">NO_ERROR</xref>.
+        </t>
+        <t>
+          If an endpoint detects multiple different errors,
+          it MAY choose to report any one of those errors. If a frame would cause a connection error, that error MUST be reported.
+        </t>
         <section anchor="ConnectionErrorHandler">
           <name>Connection Error Handling</name>
           <t>


draft-ietf-httpbis-http2bis.xml
```diff
@@ -1415,6 +1415,16 @@
         <t>
           A list of error codes is included in <xref target="ErrorCodes"/>.
         </t>
+        <t>
+          It is possible that an endpoint will encounter frames that would cause multiple errors. Implementations MAY process
+          multiple errors triggered by a single frame in any order, within the limitations set out in this section. Once an
```
martinthomson, line 1420:
```suggestion
          It is possible that an endpoint will encounter frames that would cause multiple errors. Implementations MAY discover
          multiple errors during processing, but they SHOULD report at most one stream and one connection error as a result.
```

And then I think that we need a new paragraph, though you can join the text I suggest below.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -1415,6 +1415,16 @@
...
+          implementation has handled a connection error, no further frame processing will be possible. As a result, connection
+          errors are invariably the last error type handled by an implementation.
```
martinthomson, line 1422: Unfortunately, I think that we can change our minds when we report a connection error.  And we can report stream errors after a connection error.  But we should advise against it.

> The first stream error reported for a given stream prevents any other errors on that stream from being reported.  In comparison, the protocol permits multiple GOAWAY frames, though an endpoint SHOULD report just one type of connection errors unless a graceful shutdown (see $$SECTION) is cut short as a result of a connection error is encountered.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -1415,6 +1415,16 @@
...
+        </t>
+        <t>
+          If an endpoint detects multiple stream errors caused by a single frame, but no connection errors caused by that frame,
```
martinthomson, line 1425:
```suggestion
          If an endpoint detects multiple different errors,
```

I don't think that we need to limit this to frame processing.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -1415,6 +1415,16 @@
...
+          it MAY choose to handle any one of those errors. If a frame would cause a connection error, that error MUST be handled.
```
martinthomson, line 1426:
```suggestion
          it MAY choose to report any one of those errors. If a frame would cause a connection error, that error MUST be reported.
```

martinthomson (commented):
I realize that the word "handling" is in the title, but I think that this really about reporting.  "Handling" implies that the errors are somehow dealt with, rather than having them cause stream or connection termination.

I'm thinking we can change the title to "Error Reporting" on that basis.

Lukasa:
Thanks @martinthomson, ready for re-review I think.

Lukasa:
Oh, you proposed changing "handling" to "reporting": did you want to do that pervasively in this document?

martinthomson:
Let's take that up separately.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -1415,6 +1415,21 @@
...
+        <t>
+          The first stream error reported for a given stream prevents any other errors on that stream from being reported.
+          In comparison, the protocol permits multiple <xref target="GOAWAY" format="none">GOAWAY</xref> frames, though an
+          endpoint SHOULD report just one type of connection error unless an error is encountered during graceful shutdown.
+          If this occurs, an endpoint MAY send an additional GOAWAY frame with the new error code, in addition to any prior
+          GOAWAY that contained <xref target="NO_ERROR" format="none">NO_ERROR</xref>.
+        </t>
+        <t>
+          If an endpoint detects multiple different errors,
+          it MAY choose to report any one of those errors. If a frame would cause a connection error, that error MUST be reported.
```
martinthomson, line 1431:
```suggestion
          it MAY choose to report any one of those errors. If a frame causes a connection error, that error MUST be reported.
```

martinthomson (approved):
Thanks!

Lukasa:
Ready for merge I think.

martinthomson:
I agree, but as this is substantive, let's run this by the working group.

martinthomson:
As there is a draft deadline, I will open an issue for my niggle, and ask for forgiveness rather than permission.

PR #820, opened 2021-02-17
Remove Upgrade mechanism.

### LINKED ISSUE #772: Remove upgrade mechanism
This has not been implemented, so can likely go.

See the discussion on https://github.com/martinthomson/http2v2/issues/2 for more context.

Lukasa:
HTTP/1.1 to HTTP/2 Upgrade was never widely deployed and has no
particular defenders. This change removes the text explaining and
defining the mechanism.

Resolves #772.

This change still requires discussion on the mailing list.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -299,7 +299,7 @@
             </t>
             <t>
                 The "h2c" string was previously used as a token for use in the HTTP Upgrade mechanism's
-                Upgrade header field (<xref target="RFC7230" section="6.7"/>). This usage was never widely
+                Upgrade header field (<xref target="HTTP" section="7.8"/>). This usage was never widely
                 deployed, and is no longer specified in this document.
             </t>
           </li>
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -268,8 +268,9 @@
       </t>
       <t>
         The means by which support for HTTP/2 is determined is different for "http" and "https"
-        URIs. Discovery for "http" URIs is described in <xref target="discover-http"/>.  Discovery
-        for "https" URIs is described in <xref target="discover-https"/>.
+        URIs.  Discovery for "https" URIs is described in <xref target="discover-https"/>. HTTP/2
+        support for "http" URIs can only be discovered by out-of-band means, and requires prior knowledge
+        of the support as described in <xref target="known-http"/>.
       </t>
       <section anchor="versioning">
         <name>HTTP/2 Version Identification</name>
@@ -290,12 +291,16 @@
           <li>
             <t>
                 The string "h2c" identifies the protocol where HTTP/2 is run over cleartext TCP.
-                This identifier is used in the HTTP/1.1 Upgrade header field and in any place where
-                HTTP/2 over TCP is identified.
+                This identifier is used in any place where HTTP/2 over TCP is identified.
             </t>
             <t>
                 The "h2c" string is reserved from the ALPN identifier space but describes a
                 protocol that does not use TLS.
+            </t>
+            <t>
+                The "h2c" string was previously used as a token for use in the HTTP Upgrade mechanism's
+                Upgrade header field (<xref target="RFC7230" section="6.7"/>). This usage was never widely
+                deployed, and is no longer specified in this document.
             </t>
           </li>
         </ul>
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2bis.xml
```diff
@@ -263,8 +263,9 @@
       </t>
       <t>
         The means by which support for HTTP/2 is determined is different for "http" and "https"
-        URIs. Discovery for "http" URIs is described in <xref target="discover-http"/>.  Discovery
-        for "https" URIs is described in <xref target="discover-https"/>.
+        URIs.  Discovery for "https" URIs is described in <xref target="discover-https"/>. HTTP/2
```
martinthomson, line 266: Probably something for a different change, but "discovery" isn't really right.  The use of HTTP/2 for "https" is negotiated.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -464,11 +361,8 @@
           That is, the connection preface starts with the string <tt>PRI *
           HTTP/2.0\r\n\r\nSM\r\n\r\n</tt>. This sequence
           MUST be followed by a <xref target="SETTINGS" format="none">SETTINGS</xref> frame (<xref target="SETTINGS"/>), which
-          MAY be empty. The client sends the client connection preface immediately upon receipt of
-          a 101 (Switching Protocols) response (indicating a successful upgrade) or as the first
-          application data octets of a TLS connection. If starting an HTTP/2 connection with prior
-          knowledge of server support for the protocol, the client connection preface is sent upon
-          connection establishment.
+          MAY be empty. The client sends the client connection preface as the first
+          application data octets of a TLS connection.
```
martinthomson, line 365:
```suggestion
          application data octets of a connection.
```

Might as well make the statement general :)

draft-ietf-httpbis-http2bis.xml
```diff
@@ -4271,38 +4157,6 @@
           </tbody>
         </table>
       </section>
-      <section>
-        <name>HTTP2-Settings Header Field Registration</name>
```
martinthomson, line 4275: I think that Alexey's point here about marking this obsoleted is worthwhile.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -4333,30 +4187,6 @@
             </dd>
         </dl>
       </section>
-      <section anchor="iana-h2c">
-        <name>The h2c Upgrade Token</name>
```
martinthomson, line 4337: As this registry doesn't have a status column, we can leave this untouched, with the reference to 7540 left alone.

martinthomson (approved):

Lukasa:
@martinthomson I've returned the two IANA registry entries and pointed them backwards at RFC 7540, as the sections that define those fields were removed from this document. I'm not actually sure that's the right IETF process there. Happy to take guidance as to what I should be doing instead (e.g. putting in a pair of dummy sections saying that the field and token don't have semantics any longer and referring to those instead).

draft-ietf-httpbis-http2bis.xml
```diff
@@ -4288,18 +4174,19 @@
             </dd>
```
martinthomson, line 4174: Up there^^ probably best to say that this updates it to obsolete it.  I think that you could just say something like "this marks the header field as as obsoleted" and avoid the registration block.  A backward reference to its original definition is probably all you need in addition to that.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -4353,7 +4240,7 @@
              None
```
martinthomson, line 4240: This section can probably be a tombstone only.  "Previous versions (see X of 7540) registered an upgrade token, but this capability has been removed; see Section X."

Lukasa:
Ok, I've replaced the registrations with tombstones, this is ready for another go around.

martinthomson (approved):

bencebeky (approved):

ianswett (approved):

mnot:
It would be good to more explicitly say that the upgrade mechanism is NOT RECOMMENDED, both because we're encrypting everything and because of the [new attack vectors](https://labs.bishopfox.com/tech-blog/h2c-smuggling-request-smuggling-via-http/2-cleartext-h2c) it enables (even if they're mostly shoddy implementation, it's more surface area).

martinthomson:
Sounds like a reasonable addition, but it might be better to track that with a separate change.

Lukasa:
Happy to add that here or make a separate change, either way.

martinthomson:
Looking into this a little more, I think that Section 10.3 ([Intermediary Encapsulation Attacks](https://httpwg.org/http2-spec/draft-ietf-httpbis-http2bis.html#name-intermediary-encapsulation-)) is probably where some extra warnings might be worthwhile.

The reason I might prefer a separate change is that we can hit Upgrade, Transfer-Encoding, Content-Length, and other Connection fields at the same time if we do it right.

PR #821, opened 2021-02-18
Speak more about DoS

### LINKED ISSUE #775: Improve resource exhaustion attack text
The spec includes language on this, but it turned out that many implementations failed to properly safeguard themselves. Clearer descriptions of problems based on this experience would be good.

martinthomson:
After carefully reviewing the Netflix advisories, it was clear that the
problem was that implementations didn't read the original text.

The TCP window reduction attack in CVE-2019-9517 was not mentioned, and
it is a clever exploit.  The rest were clearly already documented.

I took the opportunity to restructure things a little bit.  The first
few CVEs were about causing the victim to enqueue a bunch of frames and
then exploit the fact that this queuing was badly inefficient.  Those
got a list of their own.

I purposefully left CVE-2019-9516 out.  Though I am sad that people
didn't free() after malloc(), this is pretty much just a systematized
memory leak and I don't feel like calling that out in an RFC.  I've
cited the disclosure page, so I think we're covered there anyway.

Closes #775.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -3717,49 +3717,96 @@
           features ensure that memory commitments for these features are strictly bounded.
         </t>
         <t>
-          The number of <xref target="PUSH_PROMISE" format="none">PUSH_PROMISE</xref> frames is not constrained in the same fashion.
-          A client that accepts server push SHOULD limit the number of streams it allows to be in
-          the "reserved (remote)" state.  An excessive number of server push streams can be treated as
-          a <xref target="StreamErrorHandler">stream error</xref> of type
-          <xref target="ENHANCE_YOUR_CALM" format="none">ENHANCE_YOUR_CALM</xref>.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2bis.xml
```diff
@@ -3717,49 +3717,96 @@
...
-          known to clients and could be exceeded without being an obvious protocol violation.
+          A number of HTTP/2 implementations were found to be vulnerable to denial of service <xref
+          target="NFLX-2019-002"/>.  The following lists known ways that implementations might be
+          subject to denial of service attack:
         </t>
+
+        <ul>
+          <li>
+            <t>
+              Inefficient tracking of outstanding frames can lead to overload if an adversary can
```
Lukasa, line 3736:
```suggestion
              Inefficient tracking of outstanding outbound frames can lead to overload if an adversary can
```

Bit of a nit here but it might be helpful to be as explicit as possible.

Lukasa (approved):
One minor suggestion, otherwise I’m happy with the new text.

PR #823, opened 2021-02-18
Hervé was the only one

### LINKED ISSUE #791: Use Unicode for names
There are a number of places where we made some awkward compromises when it came to representing names.  We can undo those now.

martinthomson:
That is, in text we added.

I have an open question to rfc-interest@ regarding correcting the spelling of author names in references.  It's pretty clear [what Bodo thinks](https://www.bmoeller.de/) about the mangling we have done to his name in pursuit of ASCII and I'd like to correct that.  I didn't see any others that would appear in the final document (Hervé is there too, but only first initials survive).

Closes #791.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -5099,7 +5099,7 @@
           </li>
         <li>
             Mark Nottingham, Julian Reschke, James Snell, Jeff Pinner, Mike Bishop,
-            and Herve Ruellan (Substantial editorial contributions).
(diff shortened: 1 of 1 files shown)


Lukasa (approved):
I think making this right is a good idea, notwithstanding the questions about the pursuit of ASCII. It’d be nice if the IETF could make its way into 1995.

martinthomson:
Based on feedback from the TRSE, I'm not going to update other names.  Referential integrity is - apparently - more important.

PR #825, opened 2021-02-23
Update of "Prohibited TLS 1.2 Cipher Suites"

emanjon:
RFC 7540 is an awesome TLS 1.2 profile, a often recomment it for that purpose. 

As far as I know, there has been 4 new ciphersuites (2 for TLS 1.2 and 2 for TLS 1.3) and 1 PskKeyExchangeMode violating the null and ephemeral key exchange requirements mandated by RFC 7540.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -4835,7 +4835,7 @@
       </references>
     </references>
     <section anchor="BadCipherSuites">
-      <name>Prohibited TLS 1.2 Cipher Suites</name>
+      <name>Prohibited TLS 1.2 Cipher Suites and PskKeyExchangeModes</name>
       <t>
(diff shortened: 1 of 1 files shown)


martinthomson (commented):
Unfortunately, we cannot add to the list of prohibited cipher suites without also defining a new ALPN.  A client that is aware of these new requirements and forced to offer these suites with potential use with non-h2 ALPN values cannot assume that a server is also aware of these new requirements.  If the server is only compliant with RFC 7540 it won't know that these suites are bad (well, aside from that being obvious).  A client can only use INADEQUATE_SECURITY if it knows that the server knows that the choice of cipher suite is wrong.

We knew this to be a limitation of the profile when we built it.  That's why it says:

> Additional cipher suites with these properties could be defined; these would not be explicitly prohibited.

PR #830, opened 2021-03-22
Rewrite :authority description

### LINKED ISSUE #828: incorrect requirements for pseudo header field mapping for asterisk form
I'm struggling with <https://greenbytes.de/tech/webdav/rfc7540.html#HttpRequest>:

> The :authority pseudo-header field includes the authority portion of the target URI ([RFC3986], Section 3.2). The authority MUST NOT include the deprecated userinfo subcomponent for http or https schemed URIs.
>
> To ensure that the HTTP/1.1 request line can be reproduced accurately, this pseudo-header field MUST be omitted when translating from an HTTP/1.1 request that has a request target in origin or asterisk form (see [RFC7230], Section 5.3).

If I have

~~~
  OPTIONS * HTTP/1.1
  Host: example.com
~~~

and wanted to map this to H2, why would I omit ":authority" here?

Wild guess: the spec should say "origin or authority form", not "origin or asterisk form"?

martinthomson:
I think that this is better, but I'd like input on that.

I've tried to simplify this into 3 chunks:

* requests generated natively in HTTP/2
* translating HTTP/1.1 -> HTTP/2
* translating HTTP/2 -> HTTP/1.1

I realized that absolute-form was underserved here.  I could have
excluded it on the basis that it is special, but the rules seemed fairly
clear - at least regarding :authority, so that went in too.

The text now deals more fully with absolute form.  The previous text
really only hinted at the possibility that there might be two values.
This points to -messaging for what to do about those two values (hint:
:authority wins).

Closes #828.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -2926,15 +2926,28 @@
                     <tt>http</tt> or <tt>https</tt> schemed URIs.
                 </t>
                 <t>
-                    To ensure that HTTP/1.1 control data can be reproduced accurately, this
-                    pseudo-header field MUST be omitted when translating from an HTTP/1.1 request
-                    that has a request target in origin or asterisk form (see
-                    <xref target="HTTP11" section="3.2"/>). Clients that generate HTTP/2 requests
-                    directly SHOULD use the <tt>:authority</tt> pseudo-header
-                    field instead of the <tt>Host</tt> header field. An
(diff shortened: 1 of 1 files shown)


reschke (commented):
This looks good overall.

As a separate task once the changes are in I would recommend considering to have obe sub section per pseudo header field (and maybe combine request/response behaviour in the same place).

Lukasa (approved):
LGTM. I agree with @reschke that separate sections for these pseudo-headers would be good.

PR #832, opened 2021-03-22
:scheme is just the scheme

### LINKED ISSUE #831: Setting :scheme for OPTIONS *
See the discussion in #828 and httpwg/http-core#801.

We need to explain what to put here.  We never said anything before.

martinthomson:
This is either taken from the target URI, or translated in (S3.3 of
-messaging).  Unless it is CONNECT where is is absent.

Closes #831.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -2911,7 +2911,7 @@
               <li>
                 <t>
                   The <tt>:scheme</tt> pseudo-header field includes the scheme portion of the
-                  request.  The scheme is taken from the target URI (<xref target="RFC3986"
+                  request target.  The scheme is taken from the target URI (<xref target="RFC3986"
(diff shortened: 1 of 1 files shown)


Lukasa (approved):
Seems reasonable to me.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -2910,12 +2910,16 @@
               </li>
               <li>
                 <t>
-                    The <tt>:scheme</tt> pseudo-header field includes the scheme
-                    portion of the target URI (<xref target="RFC3986" section="3.1"/>).
+                  The <tt>:scheme</tt> pseudo-header field includes the scheme portion of the
+                  request.  The scheme is taken from the target URI (<xref target="RFC3986"
```
martinthomson, line 2914:
```suggestion
                  request target.  The scheme is taken from the target URI (<xref target="RFC3986"
```

Lukasa (approved):

PR #833, opened 2021-03-22
SETTINGS_ENABLE_PUSH from the server

### LINKED ISSUE #793: ENABLE_PUSH from clients
The definition of ENABLE_PUSH implies that it can be set to 0 by a server, but the effect of that is meaningless as it results in treating PUSH_PROMISE as an error.  PUSH_PROMISE can't be sent by a client.

Should we update the definition?

martinthomson:
The previous text implied that a value of 1 was a problem,
which was somewhat in tension with the default being 1 and servers
mostly not sending the setting.  So clearly what was being checked by
client was the frame, not the value.  This changes to checking the
setting as it arrives, changes where that text lives.

Also, this changes from "endpoint" to "client" or "server" as
appropriate to match actual practice.

Closes #793.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -2014,12 +2014,11 @@
                 target="PROTOCOL_ERROR" format="none">PROTOCOL_ERROR</xref>.
               </t>
               <t>
-                A server MUST NOT explicitly set this value to 1.  A server MAY choose to always
-                omit this setting when it sends a SETTINGS frame, but if a server does include a
-                value it MUST be 0.  A client MUST treat receipt of a SETTINGS frame with
-                SETTINGS_ENABLE_PUSH set to 1 as a <xref target="ConnectionErrorHandler">connection
-                error</xref> of type <xref target="PROTOCOL_ERROR"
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2bis.xml
```diff
@@ -1999,17 +1999,27 @@
...
+                The initial value of SETTINGS_ENABLE_PUSH is 1, which indicates that server push is
+                permitted.  Any value other than 0 or 1 MUST be treated as a <xref
+                target="ConnectionErrorHandler">connection error</xref> of type <xref
+                target="PROTOCOL_ERROR" format="none">PROTOCOL_ERROR</xref>.
               </t>
               <t>
-                  The initial value is 1, which indicates that server push is permitted.  Any value
-                  other than 0 or 1 MUST be treated as a <xref target="ConnectionErrorHandler">connection error</xref> of type
-                  <xref target="PROTOCOL_ERROR" format="none">PROTOCOL_ERROR</xref>.
+                A server MUST NOT explicitly set this value to 1.  A server MAY choose to always
```
Lukasa, line 2017: Nit: I think "always" is redundant here.

Lukasa (approved):
One minor language nit, otherwise seems great.

PR #835, opened 2021-04-01
Server Push caveats

### LINKED ISSUE #786: Cut Server Push
@ianswett says:

It's not very widely used and many implementations either don't support it at all or don't support it in a way that improves end-user performance.

This sparked a lively discussion at https://github.com/martinthomson/http2v2/issues/16.

mnot:
Fixes #786

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -137,8 +137,8 @@
         requests to be compressed into one packet.
       </t>
       <t>
-        HTTP/2 adds a new interaction mode whereby a server can <xref target="PushResources">push
-        responses to a client</xref>.  Server push allows a server to speculatively send data to a
+        Finally, HTTP/2 adds a new, optional interaction mode whereby a server can <xref target="PushResources">push
+        responses to a client</xref>.  This is intended to allow a server to speculatively send data to a
         client that the server anticipates the client will need, trading off some network usage
         against a potential latency gain.  The server does this by synthesizing a request, which it
         sends as a <xref target="PUSH_PROMISE" format="none">PUSH_PROMISE</xref> frame.  The server is then able to send a response to
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -3218,17 +3218,20 @@
           client-initiated request.
         </t>
         <t>
-          Server push was designed to allow a server to exploit knowledge about the resources it
-          serves, along with typical client behavior, to improve client-perceived performance, by
-          reducing the latency associated with content that links to other content that needs to be
-          fetched to present it (for example, stylesheets and scripts, in the case of HTML).
+          Server push was designed to allow a server to improve client-perceived performance by
+          predicting what requests will follow those that it receives, thereby removing a round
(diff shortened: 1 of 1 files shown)


Lukasa (approved):
LGTM

draft-ietf-httpbis-http2bis.xml
```diff
@@ -3216,9 +3215,20 @@
...
-          client-initiated request. This can be useful when the server knows the client will need
-          to have those responses available in order to fully process the response to the original
-          request.
+          client-initiated request.
+        </t>
+        <t>
+          Server push was designed to allow a server to exploit knowledge about the resources it
+          serves, along with typical client behavior, to improve client-perceived performance, by
+          reducing the latency associated with content that links to other content that needs to be
+          fetched to present it (for example, stylesheets and scripts, in the case of HTML).
```
martinthomson, line 3224: This is one mega-sentence.  How about:

```suggestion
          Server push was designed to allow a server to exploit knowledge about resources to
          improve client-perceived performance.  A server that could correctly predict a request
          might reduce latency by pushing a response to clients. For example, with HTML, server
          push might be used for linked content like stylesheets, scripts, and images.
```

martinthomson (approved):
This is great.  I had some trouble with the mega-sentence.  That said, my rewording might not be much better.

draft-ietf-httpbis-http2bis.xml
mnot: PTAL at latest commit

draft-ietf-httpbis-http2bis.xml
```diff
@@ -3216,9 +3215,23 @@
...
-          client-initiated request. This can be useful when the server knows the client will need
-          to have those responses available in order to fully process the response to the original
-          request.
+          client-initiated request.
+        </t>
+        <t>
+          Server push was designed to allow a server to improve client-perceived performance by
+          predicting what requests will follow those that it receives, thereby removing a round
+          trip for them. For example, a request for HTML is often followed by requests
+          for stylesheets and scripts associated with that page. When these requests
```
martinthomson, line 3224:
```suggestion
          for stylesheets and scripts referenced by that page. When these requests
```

draft-ietf-httpbis-http2bis.xml
```diff
@@ -3216,9 +3215,23 @@
...
+          are pushed, the client does not need to wait to receive the references to them in the HTML
+          and issue separate requests.
+        </t>
+        <t>
+          In practice, server push is difficult to use effectively, because it requires the
+          server to correctly anticipate the additional requests the client will make, taking into
+          account factors such as caching, content negotiation, and user behavior. Errors in
+          prediction can lead to performance degradation, due to the opportunity cost that the
+          additional data on the wire represents. In particular, pushing more than a round trip of
+          data can cause contention issues with more-important responses.
```
martinthomson, line 3234: I think that you might say "pushing any significant amount of data".  It doesn't need to be a whole round trip worth.  Even a fraction of a round trip worth can starve out a high priority request that arrives immediately after starting.

martinthomson (approved):
Very nice.  Thanks Mark.

PR #837, opened 2021-04-22
Stop using "settings parameters"

martinthomson:
In the spirit of the naming simplications of the recent HTTP changes, we
can talk about settings in this revision.

This aligns better with HTTP/3, though that still uses "settings
parameters" (with that ugly double-plural).

cc @MikeBishop


Lukasa (approved):

MikeBishop:
I mean, it's not too late to Auth48 the same change into H3....  Not quite, anyway.

PR #844, opened 2021-04-23
Do the obsoletes thing properly

### LINKED ISSUE #842: should reference RFCs 7540 and 8740
informatively, as they are obsoleted by this spec

martinthomson:
The current rules state that you need to include text in abstract and
introduction.

This will result in redundant entries for the specs with other changes that add 7540 references, but we can sort that out later. 

Closes #842.


Lukasa (approved):
Rules are boring. Be a rebel without a pause. Omit that text.

martinthomson:
Picking my battles more judiciously (as much as I'd like to fight this one, it's a low reward, high risk situation).

PR #845, opened 2021-04-23
Make HTTP/1.1 reference informational

### LINKED ISSUE #841: Tweak HTTP/1.1 dependencies
Right now, we have a lot of 1.1-specific language.  In theory, it should be possible to remove 1.1 references and use core HTTP semantics to describe the protocol elements.  That would be much cleaner, but it could be tricky.

martinthomson:
This required a little careful rewriting of text that only recently
rewrote.  Hopefully this retains the improvements of that rewrite.

Closes #841.


Lukasa (approved):
Nice. I think the retains the original meaning of the text well enough, and is as clear as it can be without ever needing to normatively invoke HTTP/1.1. Needle well threaded.

PR #848, opened 2021-04-27
xml2rfc workaround

martinthomson:
xml2rfc is overly pedantic about non-ASCII text.  The result in this
case was that Hervé's name was being mangled to include a literal
`&#233;`, even in a text format (where HTML character references make no
sense).  To avoid that, this is what is necessary.  Of course, this
makes the rendering of the document with the RFC Editor stylesheet look
awful.  I've decided not to concern myself with that part though; the
stylesheet we use is fine.


Lukasa (approved):
Good to see the IETF is fully unicode aware. 😐

PR #850, opened 2021-05-24
Clean closed state

### LINKED ISSUE #783: Try to improve the definition of the closed state
Rules around the closed state are particularly ambigous, because we lack a "closing" or equivalent state and any forced local close by RST_STREAM results in incoming frames violating the rules for what is acceptable in closed state since the other end doesn't know it's closed. It used to be even more problematic with priorities since some dependencies could be completely changed. I had a proposal to improve this using a form of ACK for RST/GOAWAY using PING that would allow one side to know when the other side was notified of these frames, but with priorities gone I don't think that's needed anymore. However we still need to be careful about the MUST NOT wording around whatever involves the closed state.

Suggested by @wtarreau.

martinthomson:
Closes #783.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -974,54 +974,59 @@
           <dt>closed:</dt>
           <dd>
             <t>
+              The "closed" state is the terminal state.
+            </t>
+            <t>
+              An stream enters the "closed" state after an endpoint both sends and receives a frame
+              with an END_STREAM flag set. A stream also enters the "closed" state after an endpoint
+              either sends or receives a <xref target="RST_STREAM" format="none">RST_STREAM</xref>
+              frame.
+            </t>
+            <t>
+              An endpoint MUST NOT send frames other than <xref target="PRIORITY"
+              format="none">PRIORITY</xref> on a closed stream.  An endpoint MAY treat receipt of
+              any other type of frame on a "closed" stream as a <xref
+              target="ConnectionErrorHandler">connection error</xref> of type <xref
+              target="STREAM_CLOSED" format="none">STREAM_CLOSED</xref>, except as noted below.
+            </t>
+            <t>
+              An endpoint that sends a frame with the END_STREAM flag set or a <xref
+              target="RST_STREAM" format="none">RST_STREAM</xref> frame might receive a <xref
+              target="WINDOW_UPDATE" format="none">WINDOW_UPDATE</xref> or <xref target="RST_STREAM"
+              format="none">RST_STREAM</xref> frame from its peer in the time before the peer
+              receives and processes the frame that closes the stream.
+            </t>
+            <t>
+              An endpoint that sends a <xref target="RST_STREAM" format="none">RST_STREAM</xref>
+              frame on a stream that is in the "open" state could receive any type of frame.  The
+              peer might have sent or enqueued for sending these frames before processing the <xref
+              target="RST_STREAM" format="none">RST_STREAM</xref> frame. An endpoint MUST minimally
+              process and then discard any frames it receives in this state.  This means updating
+              header compression state for <xref target="HEADERS" format="none">HEADERS</xref> and
+              <xref target="PUSH_PROMISE" format="none">PUSH_PROMISE</xref> frames; <xref
+              target="PUSH_PROMISE" format="none">PUSH_PROMISE</xref> frames also cause the promised
+              stream to become "reserved", even when the <xref target="PUSH_PROMISE"
+              format="none">PUSH_PROMISE</xref> frame is received on a closed stream; and, the
+              content of <xref target="DATA" format="none">DATA</xref> frames counts toward the
+              connection flow-control window.
+            </t>
+            <t>
+              An endpoint can perform this minimal processing for all streams that are in the
+              "closed" state.  Endpoints MAY use other signals to detect that a peer has received
+              the frames that caused stream to become "closed" and treat receipt of any frame other
+              than <xref target="PRIORITY" format="none">PRIORITY</xref> as a <xref
+              target="ConnectionErrorHandler">connection error</xref> of type <xref
+              target="PROTOCOL_ERROR" format="none">PROTOCOL_ERROR</xref>.  Endpoints can use frames
+              that indicate that the peer has received the closing signal to drive this.  Endpoints
+              SHOULD NOT use timers for this purpose.  For example, an endpoint that sends a <xref
+              target="SETTINGS" format="none">SETTINGS</xref> frame after closing a stream can
+              safely treat receipt of a <xref target="DATA" format="none">DATA</xref> frame on that
+              stream as an error after receiving an acknowledgement of the settings.  Other things
+              that might be used are <xref target="PING" format="none">PING</xref> frames, receiving
+              data on streams that were created after closing the stream, or responses to requests
+              created after closing the stream.
+            </t>
 
-                The "closed" state is the terminal state.
-            </t>
-            <t>
-                An endpoint MUST NOT send frames other than <xref target="PRIORITY" format="none">PRIORITY</xref> on a closed
-                stream.  An endpoint that receives any frame other than <xref target="PRIORITY" format="none">PRIORITY</xref>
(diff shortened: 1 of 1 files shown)


wtarreau:
Thanks Martin for working on this!
Regarding the following:
> Endpoints  SHOULD NOT use timers for this purpose.

I'd add "unless they have a way to know the timer is at least as large as the connection's RTT". While I'm not using timers myself, I can easily imagine that for some implementations that permanently monitor their connections' quality of service it could be safe and easy to use timers.

After the list of proposed methods, I would add the stream counter one because I think it will often be the easiest one to implement: "seeing at least SETTINGS_MAX_CONCURRENT_STREAMS new streams created since the RST_STREAM frame was sent also proves it was received and processed".

What I'm unsure about is whether we should explicitly request for implementations to be careful not to send extra RST_STREAM during this period or if we should only rely on other mentions for this, or leave it implied by "no frame other than ...".

martinthomson:
The timers thing is something that I realized might run afoul of certain practices that we know to be common.  Let's say that you have large send buffers (for performance) and you enqueue more than a BDP.  In that case, if the peer resets the stream, you might have committed to writing so much that even after a few round trips, the buffer is still draining.

Now maybe that is enough cause to drop the connection, but it's a practice that isn't technically *wrong*.  Enforcing this on a timer makes that more risky.

As for recycling the entire stream limit, I don't know if it is true that it is easy to implement.  There are obvious ways to implement it that are easy but also obviously wrong.  I would prefer not to include it.

I'm not following your point about extra RST_STREAM frames.  Do you mean sending RST_STREAM for the same stream more than once?  There is no point in doing that and I think we already prohibit that.  If an endpoint has to reset other streams, I don't think that we can stop that.

wtarreau:
I agree that timers are far from being perfect especially in window-in-window environments like H2, but every implementation has some at least for communication timeouts. In addition, in the case you mention about huge data being queued, that's true but in such a case each such data has to reset the timer. And I also agree that extra data from other streams could artificially delay the resetted stream's final data (but that implicitly increases the measured RTT as well). I don't think we should discourage implementations from doing this if that's easier for them and they can get it right (don't forget that often the alternative is to flood with RST_STREAM in the other direction for each and every frame receivd in closed state). But you're probably right that we should not encourage it either because it's easy to get it wrong. Maybe the only thing to do is to turn the "SHOULD NOT" into a non-normative "should not".

Regarding the recycling of the streams, yes it's easy to implement. You just need a counter, which will work exactly like a timeout except that it's based on stream counts. So when resetting a stream we'd just have to set its expiration value to current stream count plus SETTINGS_MAX_CONCURRENT_STREAMS and you're sure that the other end has necessarily received your reset otherwise it would be blocked at a lower value. And I'm pretty sure that it is what I'll implement in the end (much easier and cleaner than sending a PING frame after each RST_STREAM as I initially thought about).

Regarding the extra RST_STREAM frames, yes that's what I mean. We do prohibit that but we offer no easy way to help against doing it, except in this discussion, which is exactly the point :-)  The problem starts by the spec explicitly asking not to send anything other than PRIORITY on a closed stream and suggesting to use RST_STREAM to close a stream. But the reality in field is that a closed stream is a stream that doesn't exist anymore (i.e. not attached to any upper application layer) and generally the only possible response to this is to send RST_STREAM to signal that the stream is closed (analogy to the "broken pipe" between processes or simply TCP RST). I'm pretty sure we fell into that situation in the initial implementation we had in haproxy. RST_STREAM should be seen more as an "abort" signal but given that it's often used to report processing errors, it's natural to use it as well to signal that the stream targetted by a received frame has just disappeared.

Often I've been wondering if it wouldn't have been easier to refer as "non-existing stream" or "unknown stream" for those that have been definitely closed, and differentiate them from "closed" ones which are still known. We could then say that a closed stream disappears immediately on END_STREAM+half-closed and on receipt of RST_STREAM, but that it should be kept in closed state for some time after sending RST_STREAM. Then instead of speaking about "stream closed by RST_STREAM" or other variants we could simply speak about "closed stream" versus "unknown stream" and have simple rules for the latter, especially since the rules are the same for frames received on these closed streams and those received on idle streams. But I'm a bit wary of revisiting all places mentioning "closed" :-/

martinthomson:
OK, that's a bunch of stuff to unpack.

On the timer issue, I think that you just convinced me that dissuading people from using timers is the right idea.  It's not a hard rule, so people can use timers if they want.  I could just remove the text about timers, but it seems to me like people will try to use timers and fail majestically.  The text necessary to explain the pitfalls would be more involved than I think the document needs.

On the stream limit thing, I'm convinced by the potential for it to be easy to implement (it's more state than I would keep, but that's your call).  However, explaining both how to do that AND explaining why this is OK to do would also take a lot of words.  You will note that the other options we present are pretty short (send SETTINGS, receive an ACK; send PING, receive an ACK) and they are pretty obvious.  The logic here is less direct.

On the multiple RST_STREAM thing, could you open an issue?  If people are sending multiple RST_STREAM frames for the same stream and that isn't causing problems, AND there is a good reason to do it, THEN we should document that properly.

wtarreau:
For timers, I think that just lowercasing "should-not" will be sufficient, no need for more explanation.
For the streams count, I don't think we need the put the demonstration, simply saying that "or at least SETTINGS_MAX_CONCURRENT_STREAMS new streams were initiated by the peer" is sufficient, it's left as an exercice to the reader to verify why if they want.

For the RST_STREAM one I could open an issue, it's just still unclear to me how to address it other than by this precise one. So I think that the current issue can magically solve plenty of small ones because it fixes the root cause instead of addressing individual side effects. Maybe Cory has an opinion about it ?

Lukasa:
Regarding multiple RST_STREAM I think there are only two states: either you have sent an RST_STREAM and you don't think your peer has seen it yet, in which case you should ignore frames sent on that stream, or you have sent it and you think your peer _has_ seen it but is still sending non-PRIORITY frames for that stream, in which case you should connection error with PROTOCOL_ERROR.

wtarreau:
Actually there's a third state: the stream might have reached the closed state from your peer sending an RST_STREAM itself, and the spec says that in this case you should treat any frame other than PRIORITY with a stream error (hence return an RST_STREAM). By the way this even opens the way to an RST_STREAM loop but I believe it's discussed elsewhere in the spec.

And that's precisely the problem with this "closed" state, it's that the rules to process frames in this state differ depending on how you reached that state, and not exclusively on the state. This basically means there are multiple closed states. But once you delete your stream and consider it's closed, it's not possible anymore to figure what happened.

So when you receive frames for a closed stream, you have three options:
  - you assume the peer sent a frame after END_STREAM, it's a protocol error and you break the connection
  - you assume the peer sent a frame after RST_STREAM, it's a tream error and you send RST_STREAM(stream_closed) (and in my opinion don't do that if the frame itself was RST_STREAM)
  - or you consider that you might have closed it yourself and it would be bad to punish your peer for not having received it yet, and thus you assume you were the one having sent the reset and ignore the frames.

I used to send RST_STREAM(stream_closed) when getting frames in this state because it was the most natural and efficient thing to do, with the least impact. But it can result in RST_STREAM floods. Now I prefer to simply ignore frames, but it also means it's not really possible to abort an upload with an early response (e.g. 401 or 302 on a POST) if you have to drain all input.

Lukasa:
> Actually there's a third state: the stream might have reached the closed state from your peer sending an RST_STREAM itself, and the spec says that in this case you should treat any frame other than PRIORITY with a stream error (hence return an RST_STREAM).

Where does the spec say that's a stream error?

wtarreau:
Just found it, page 19 in 7540:
```
closed:
      The "closed" state is the terminal state.

      An endpoint MUST NOT send frames other than PRIORITY on a closed
      stream.  An endpoint that receives any frame other than PRIORITY
      after receiving a RST_STREAM MUST treat that as a stream error
      (Section 5.4.2) of type STREAM_CLOSED.
```

Lukasa:
That text is replaced by the new text in this PR.

wtarreau:
OK, I didn't notice in the diff that this part was replaced (text parts are much less readable with the XML tags around). But I think we're addressing all remaining problems then, cool!

martinthomson:
Thanks all for the feedback.  We'll need to ask others to review this carefully, but the best way to do that now is to make a new draft.

Issue #826, opened 2021-03-04
Tweaking registry policies


mnot:
There's currently a slight divergence in IANA policies between h2 and h3:

- Frame Types: IETF Review, Experimental Use (h2) / Standards Action, Specification Required (h3)
- Settings: Expert Review, Experimental Use / Standards Action, Specification Required
- Error Codes: Expert Review / Standards Action, Specification Required

If we can align these with low pain, it might be worthwhile; e.g., can we get rid of the experimental use areas and upgrade Expert Review to Specification Required?

martinthomson:
h3 requires only Expert Review for provisional registrations, which makes the two basically the same.  Do we want to add provisional registrations to h2?

The experimental space can go, I think, but it might pay to check if anyone is using it first.

mnot:
+1, I think we can advertise in the WG and maybe one or two other places and that'd be sufficient, absent pushback.

martinthomson:
There are two potential changes here: dropping experimental use and adding provisional registrations.  I think that I prefer doing just the former.  The registration policies we have are loose enough that we don't really gain much from provisional registrations, nor does it really change how things work in practice.

QUIC has the notion that we might claw back provisional registrations, but that only really exists so that we can maybe reclaim the short codepoints in the small 0x00-0x3f range, which people are already camping out in as if they own them.  HTTP/2 has a flat space, which means we probably don't need to worry as much.  Frame types are scarce, but we can maybe use settings to overload those as necessary.  And the existence of HTTP/3 should relieve a lot of pressure on HTTP/2 anyway.

mnot:
I think I agree. Happy to close with the PR above already merged.

martinthomson:
\o/ thanks.

PR #853, opened 2021-05-24
Experimental begone

martinthomson:
This is part of the fix for #826.  This part removes the experimental range for frame types and settings.

Also updates to use section references to RFC 8126, which replaces RFC 5226.


Lukasa (approved):
Nice clean change. No notes.

PR #856, opened 2021-05-24
Copy the QUIC text regarding generic errors

### LINKED ISSUE #824: Any appropriate error code, including generic ones
QUIC includes the following statement:

>  In particular, an endpoint MAY use any applicable error code when it detects an error condition; a generic error code (such as PROTOCOL_VIOLATION or INTERNAL_ERROR) can always be used in place of specific error codes.

I like this.  (No accident.  I wrote it.)

Should HTTP/2 include a similar statement?  We have PROTOCOL_ERROR, which has a definition compatible with this statement, so I guess we're already there on as far as allowing a generic code goes.  Any sense in saying it explicitly?

martinthomson:
Closes #824.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -1185,8 +1185,13 @@
           GOAWAY that contained <xref target="NO_ERROR" format="none">NO_ERROR</xref>.
         </t>
         <t>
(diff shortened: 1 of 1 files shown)


Lukasa (approved):
Reads clearly, captures the intent.

PR #857, opened 2021-05-24
Remote timing attacks are a thing

### LINKED ISSUE #785: Enabling remote timing attacks
https://portswigger.net/daily-swig/researchers-exploit-http-2-wpa3-protocols-to-stage-highly-efficient-timeless-timing-attacks describes how HTTP/2 can be used to remove some network-induced jitter from the timing measurements. If your server has timing side channels, then putting requests that can be compared in the same IP packet ensures that the server receives them at the same time. This doesn't help if timing differences are masked by server queuing, CDNs or load balancers, or response delivery, but it can help reduce measurement noise.

Consider documenting this.

martinthomson:
Not sure that this needs a citation.

Closes #785.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -3720,6 +3720,22 @@
           certain scenarios.
         </t>
       </section>
+      <section>
+        <name>Remote Timing Attacks</name>
+        <t>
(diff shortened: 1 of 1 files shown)


Lukasa (approved):
Yeah, I'm not sure what else we can say here. There are risks here, there are not really many mitigations available.

mnot (approved):

PR #858, opened 2021-05-24
Add details regarding removal of connection-specific fields.

### LINKED ISSUE #789: Consider loosening requirement for clients to reject connection-specific header fields
I have no personal knowledge of this issue, but it came up on the #whatwg IRC channel. https://megamorf.gitlab.io/2019/08/27/safari-nsposixerrordomain-100-error-with-nginx-and-apache/ describes Safari rejecting responses that contain the `upgrade` header field, while other browsers accept it. It'd be good to converge the major clients on a single behavior and, if that's not the behavior currently specified, update the spec to match.

### LINKED ISSUE #851: Intermediaries and connection-specific headers
Upgrade, Transfer-Encoding, Content-Length, Connection, and anything listed in Connection MUST NOT be forwarded without modification.  See also #789.

Lukasa:
Here is some draft text. In principle we don't need any of this at all: as this text shows, all of the relevant guidance is already present in both semantics and H2. I've elected to throw up some possible text, but if we wanted to say that some or all of it is entirely redundant with existing specifications then I'm happy to do so.

Fixes #789.
Fixes #851.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -3474,6 +3474,13 @@
           an attacker if they are translated verbatim.  Any request or response that contains a
           character not permitted in a field value MUST be treated as <xref target="malformed">malformed</xref>.  Valid characters are defined by the <tt>field-content</tt> ABNF rule in <xref target="HTTP" section="5.5"/>.
         </t>
+        <t>
+          An intermediary that receives any field that requires removal before forwarding
+          (see <xref target="HTTP" section="7.6.1" />) MUST remove or replace those header fields when
+          forwarding messages. Additionally, intermediaries should take care when forwarding messages
+          containing Content-Length fields to ensure that the message is <xref target="malformed">well-formed</xref>.
+          This ensures that if the message is translated into HTTP/1.1 at any point the framing will be correct.
+        </t>
       </section>
       <section>
         <name>Cacheability of Pushed Responses</name>


draft-ietf-httpbis-http2bis.xml
```diff
@@ -3474,6 +3474,13 @@
           an attacker if they are translated verbatim.  Any request or response that contains a
           character not permitted in a field value MUST be treated as <xref target="malformed">malformed</xref>.  Valid characters are defined by the <tt>field-content</tt> ABNF rule in <xref target="HTTP" section="5.5"/>.
         </t>
+        <t>
+          An intermediary that receives any field that requires removal before forwarding
+          (see <xref target="HTTP" section="7.6.1" />) must remove or replace those header fields when
```
martinthomson, line 3479:
```suggestion
          (see <xref target="HTTP" section="7.6.1" />) MUST remove or replace those header fields when
```

draft-ietf-httpbis-http2bis.xml
```diff
@@ -3474,6 +3474,13 @@
...
+          forwarding messages, including between multiple messaging formats. Additionally, intermediaries
```
martinthomson, line 3480:
```suggestion
          forwarding messages. Intermediaries
```

I'm not sure that that adds anything.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -3474,6 +3474,13 @@
...
+          should take care when forwarding messages containing Content-Length fields to ensure that the
```
martinthomson, line 3481:
```suggestion
          MUST validate any Content-Length fields before forwarding to ensure that the
```

draft-ietf-httpbis-http2bis.xml
```diff
@@ -3474,6 +3474,13 @@
...
+          message is <xref target="malformed">well-formed</xref>.
```
martinthomson, line 3482: Not sure if we need to say why: HTTP/1.1 uses Content-Length for framing.

draft-ietf-httpbis-http2bis.xml
Lukasa: Hmm, I don't think that wording is acceptable: it implies that intermediaries are obligated to buffer all messages before they forward them on, which I don't think they'll do. We could say MUST validate any Content-Length fields _when_ forwarding, I suppose.

PR #859, opened 2021-05-24
Discuss flow control performance

### LINKED ISSUE #784: Speak a bit more about head-of-line blocking
H2 strongly suffers from head-of-line blocking when streams do not progress at the same spead, typically when they come from coalesced connections from different clients. Per-stream flow control is generally ineffective against this as practical window sizes matching a BDP often require impractical buffering between the intermediary and the client. As such we should mention in the recommendations that intermediaries that coalesce connections either have very large buffers or try to group streams from a same client connection together on a server connection.

Suggested by @wtarreau.

Lukasa:
Brought over the QUIC text about flow control performance.

cc @wtarreau

Resolves #784.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -1085,6 +1085,19 @@
             critical frames, such as <xref target="WINDOW_UPDATE" format="none">WINDOW_UPDATE</xref>, are not read and acted upon.
           </t>
         </section>
+        <section anchor="FlowControlPerformance">
+          <name>Flow Control Performance</name>
+          <t>
+            If an endpoint cannot ensure that its peer always has available flow control window space that is greater than the peer's
+            bandwidth-delay product on this connection, its receive throughput will be limited by HTTP/2 flow control. This will result
+            in degraded performance.
+          </t>
+          <t>
+            Sending timely <xref target="WINDOW_UPDATE" format="none">WINDOW_UPDATE</xref> frames can improve performance. Endpoints will
+            want to balance the need to improve receive throughput with the need to manage resource exhaustion risks, and should take
+            careful note of <xref target="dos"/> in defining their strategy to manage window sizes.
+          </t>
+        </section>
       </section>
 
       <section anchor="StreamPriority">


wtarreau:
Thank you Cory. After the first sentence, I'd like to add the following because I'm pretty certain it is easily overlooked, especially in gateways where there's basically no way to figure the *really* available room:
```
Note that advertising a larger stream window than the amount of storage effectively dedicated to the stream may sometimes result in head-of-line blocking situations by which incoming data for stream for which there is no more room will prevent any progress to be made on the whole connection.
```

Lukasa:
Under what circumstances can that happen?

draft-ietf-httpbis-http2bis.xml
```diff
@@ -1085,6 +1085,19 @@
...
         </section>
+        <section anchor="FlowControlPerformance">
+          <name>Flow Control Performance</name>
+          <t>
+            If an endpoint cannot ensure that its peer always has available flow control window space that is greater than the peer's
+            bandwidth-delay product on this connection, its receive throughput will be limited by HTTP/2 flow control. This will result
+            in degraded performance.
+          </t>
+          <t>
+            Sending timely <xref target="WINDOW_UPDATE" format="none">WINDOW_UPDATE</xref> frames can mitigate this bottleneck. Endpoints will
```
martinthomson, line 1096: QUIC says "improve performance", which I think is better than "mitigate this bottleneck".  bottleneck is a colloquialism that might be hard to understand, and mitigation isn't what you want (you want avoidance in that case).

wtarreau:
@Lukasa basically any time on the forward path: if you only count on the local buffer you're using to receive a stream, the window will always be ridiculously low, while you *know* that you have plenty of buffers on the other side (e.g. forwarding to H1, which is dedicated to your stream, you can count on the H1 buffers and system socket buffers). When forwarding to H2 you can theoretically at least count on min(stream_window,connection_window,buffer room+system buffers), though it can be quite hard to figure.

The problem with a multiplexed protocol (when doing fctl-over-fctl like H2 does with TCP) is that your per-stream buffers are *always* smaller than the connection-level buffers (that include the system's huge socket buffers), which are usually automatically tuned to accommodate the BDP. So in the end relying on stream buffers only implies you can never reach the BDP and will be forced to limit the single-stream bandwidth. This problem doesn't happen with other protocols like QUIC simply because there's no need in-order buffering in the system that you have to rely on.

In the end, a reasonable approach consists in having a configurable margin on top of the per-stream buffer room to figure what to advertise, but we need to warn about the risks of advertising too large a window.

Lukasa:
Hmm, sorry @wtarreau but I don't see how that relates to the original note you gave.

Where is there a head-of-line blocking issue? Having small stream window sizes prevents a single stream reaching full network performance, sure, but that isn't a head-of-line blocking problem. It's a window management problem. If you have multiple streams, all with small windows but all of which can make progress, you can in principal avoid any issues with window management.

What am I missing that brings HOL blocking into play?

wtarreau:
Sorry Cory if my explanation was not clear. What I meant is that when you start to use reasonable (i.e. small) windows for your streams, you immediately see the network performance become horrible, despite all the buffering capacity around (kernel etc), only because of the streams window. I've had users complain about 5 Mbps uploads for example with 64kB windows :-/ As such it's extremely tempting to overprovision the stream window to make it cover some of the extra buffers available around that you're almost certain exists (i.e. the output buffer for where that frame is going to be forwarded), but when starting to play with this you *risk* HOL because it's often very hard to be certain about *really* available room (i.e. if you forward that over H2 and several streams consider that same room as available etc). Window-in-window is a very well-known anti-pattern and we knew it was going to hit one way or another and we have to live with it here. For most use cases where a browser is the sole owner of a connection, it's trivial to advertise huge windows and deal with it this way (no real risk that the browser blocks itself). When coalescing multiple agents it becomes quite visible a problem. I hope I managed to make it clearer this time, otherwise do not hesitate to let me know :-)

Lukasa:
Thanks Willy, this is definitely clarifying things. So let me try to verify my understanding in bullets.

1. The default per-stream window size (64kB), and other small per-stream window sizes, has performance problems over fat links.

    This is true. It is not necessarily true if you have multiple streams in flight and your connection window is larger, but it is definitely true for single streams.
    
2. To that end it is tempting to advertise larger stream windows, to improve throughput.

3. However that window may not be available in the target logical stream, such that if the user actually consumes it your application would have to buffer it.

    Possible reasons for this include the TCP send buffer not having room, the target H2 connection window not having room, the target H2 stream window not having room, etc.

So this is definitely a resource management issue. Presumably the reason there's an HOL issue here is that, in response to needing to buffer on the outbound side, you stop reading from the H2 connection.

I can think of a few ways to manage this, but this isn't really the forum for that kind of engineering chat. :wink: Instead, I'll say that I think the current text covers this. Specifically, the text says "bigger windows improve performance, but you need to bear in mind the resource exhaustion concerns elsewhere in this document". Should we be saying anything else more specific?

wtarreau:
Well, I only partially agree, in that it's not exactly a resource exhaustion in the pure sense of the term as there can be plenty of resources available (socket buffers everywhere), but more a matter of difficulty/complexity to *reliably* measure available resources at any instant (i.e. the TCP buffer you have on one side is not necessarily dedicated to a single stream, so what is doable during a POST upload can suddenly change before you even have the option to advertise a window size change).

I agree that this could be discussed elsewhere in the doc in more details, but my point was to at least put a short warning like the one I proposed to raise awareness of implementors against the risk of pushing this setting too far because "it seems to work much better".

For the record, in haproxy what we've done to mitigate this on the download side is to avoid mixing multiple client connections over a single outgoing H2 connection. This way a slow client cannot block a fast one and it remains possible to advertise slightly larger stream windows that more realistically take into account available room on the path. For uploads, users generally consider that an uploading browser doesn't perform other operations in parallel over the same connection so it's okay to increase the advertised stream as well. But all these mitigations resulted from experience, were not obvious and will not necessarily apply to other implementations, which is why I thought a small warning could be useful.

Lukasa:
Great, so do you think the current text covers it?

```
Sending timely <xref target="WINDOW_UPDATE" format="none">WINDOW_UPDATE</xref> frames can improve performance. Endpoints will
want to balance the need to improve receive throughput with the need to manage resource exhaustion risks, and should take
careful note of <xref target="dos"/> in defining their strategy to manage window sizes.
```

wtarreau:
Looks great like this, thank you :-)

martinthomson:
Thanks for the constructive discussion.  No doubt we missed something (that's a constant risk here), but this looks like this change is a useful one.  We'll track further improvements with new issues as the opportunity arises.

PR #860, opened 2021-06-03
Use QUIC-style frame format illustrations

mnot:
I haven't adjusted for `artwork` vs. `sourcecode` until that war is over.

Note that I've explicitly added a format for the SETTINGS frame itself, which was missing.


LPardue:
+1 . This would make it easier to write extensions for both h2 and h3. Priorities draft suffers from the mixed presentation format right now.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -1360,13 +1363,11 @@
...
- +---------------+-----------------------------------------------+
- |                            Data (*)                         ...
- +---------------------------------------------------------------+
- |                           Padding (*)                       ...
- +---------------------------------------------------------------+
+Data Frame Payload {
+  [Pad Length (8)],
+  Data (..),
+  Padding (..),
+}
```
martinthomson, line 1370: If we're going full QUIC, then I think we should include the frame header in each of these frame definitions, just like QUIC ended up doing.  The advantage of that being that you can put the flag bits into the definition.  Then we can unify the descriptions of flags and fields.

Sure, that's a lot more work, but I think that it's worth doing.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -189,6 +189,10 @@
           Hexadecimal literals are prefixed with <tt>0x</tt> to distinguish them
           from decimal literals.
         </t>
+        <t>
+          This specification describes binary formats using the convention described in
+          Section 1.3 of <xref target="QUIC">RFC 9000</xref>.
```
martinthomson, line 194:
```suggestion
          <xref target="QUIC" section="1.3">RFC 9000</xref>.
```

martinthomson (commented):
Let's float this with the working group at the interim.  I don't think that this is quite ready to take, but I do like the direction it is going.

draft-ietf-httpbis-http2bis.xml
mnot: So, something like 

```
Data Frame {
  Length (24),
  Type (8),
  END_STREAM (1),
  Flag 1 (1),
  Flag 2 (1),
  PADDED (1),
  Flag 4 (1),
  Flag 5 (1),
  Flag 6 (1),
  Flag 7 (1),
  R (1),
  Stream Identifier (31),
  [Pad Length (8)],
  Data (..),
  Padding (..),
} 
```

?

martinthomson: I would probably compress the unused flags, but yeah.

```  
  END_STREAM (1),
  Unused Flags (2),
  PADDED (1),
  Unused Flags (4),
```

Or maybe
```
  END_STREAM Flag (1),
```

The main problem with this structure is that this probably needs to come with changes to the way that these are described.

mnot: Understood re: descriptions; just want to get the format convention down first.

I didn't compress unused flags because it seems like it'd be super-useful for subsequent extensions that want to define flags. Thoughts?

mnot: Now I'm wondering whether it makes sense to leave flags in a separate prose list, or move them up into the same list as the other descriptions. Got any gifs for that?

draft-ietf-httpbis-http2bis.xml
```diff
@@ -417,15 +421,14 @@
...
- +-+-------------+---------------+-------------------------------+
- |R|                 Stream Identifier (31)                      |
- +=+=============================================================+
- |                   Frame Payload (0...)                      ...
- +---------------------------------------------------------------+
+HTTP Frame {
+  Length (24),
+  Type (8),
+  Flags (8),
+  R (1),
```
Lukasa, line 428: Do we need to call this `R` in the new format? Can we use our extra horizontal space and write out the whole word "Reserved"`?

draft-ietf-httpbis-http2bis.xml
```diff
@@ -1441,24 +1451,33 @@
...
+  Unused Flag (1),
+  END_HEADERS Flag (1),
+  PADDED Flag (1),
+  Unused Flag (1),
+  PRIORITY Flag (1),
+  Unused Flags (2),
+  R (1),
+  Stream Identifier (31),
+  [Pad Length (8)],
+  E (1),
```
Lukasa, line 1470: Similarly for `E` here, can we use the word `Exclusive`?

draft-ietf-httpbis-http2bis.xml
```diff
@@ -1558,18 +1577,24 @@
...
- +-+-------------+-----------------------------------------------+
- |   Weight (8)  |
- +-+-------------+
+Priority Frame {
+  Length (24),
+  Type (8),
+  Unused Flags (8),
+  R (1),
+  Stream Identifier (31),
+  E (1),
```
Lukasa, line 1589: And here.

Lukasa (commented):
I generally like this.

draft-ietf-httpbis-http2bis.xml
bagder: Yes please, this allows for "proper" names there!

draft-ietf-httpbis-http2bis.xml
mnot: Works for me, as long as it doesn't rock anybody's world.

draft-ietf-httpbis-http2bis.xml
mnot: We already have an Exclusive Flag; having an Exclusive Field is going to confuse people. Can we come up with another name that distinguishes them a bit more?

mnot: Oh wait, no we don't -- that's just loose use of the term 'flag' in 7540. Hold on a sec...

draft-ietf-httpbis-http2bis.xml
```diff
@@ -1357,20 +1360,27 @@
...
- +---------------------------------------------------------------+
- |                           Padding (*)                       ...
- +---------------------------------------------------------------+
+Data Frame {
+  Length (24),
+  Type (8),
+  END_STREAM Flag (1),
+  Unused Flags (2),
+  PADDED Flag (1),
+  Unused Flags (4),
```
Lukasa, line 1372: I believe these flags (and all the other flags) are defined in the wrong bit order.

This is...not entirely well specified in RFC 9000, but the example given in § 1.3 is:

```
Example Structure {
  One-bit Field (1),
  7-bit Field with Fixed Value (7) = 61,
  Field with Variable-Length Integer (i),
  Arbitrary-Length Field (..),
  Variable-Length Field (8..24),
  Field With Minimum Length (16..),
  Field With Maximum Length (..128),
  [Optional Field (64)],
  Repeated Field (8) ...,
}
```

With the following prose after it:

> When a single-bit field is referenced in prose, the position of that field can be clarified by using the value of the byte that carries the field with the field's value set. For example, the value 0x80 could be used to refer to the single-bit field in the most significant bit of the byte, such as One-bit Field in Figure 1.

This implies that `0x80` refers to the first bit of a byte. `END_STREAM` is not 0x80, it's 0x1. All of the flags sections need to be inverted.

mnot: The prose uses indexed bits - e.g., for `END_STREAM` it says 'When set, bit 0...'. I had assumed that it was indexing from left, this being network byte order and all, but you could also argue it's indexing from least significant bit (although that's pretty odd IMO).

Nevertheless, this protocol interoperates somehow. 

If I read wireshark correctly, your interpretation is correct:

Screen Shot 2021-07-12 at 3 28 58 pm

However did we get this far?

mnot: I'm about to push a commit that should correct that. Please review carefully.

We also might want to reconsider the language we use about ordering...

martinthomson:
OK, thanks for doing this @mnot.  I merged this manually after fixing a few things up.  From my commit:

*    Use all caps for frame names
*    Add a value for Type in frame definitions
*    Change order of flag descriptions to match order in layout
*    Make Exclusive field of PRIORITY optional
*    Move PING ACK Flag to the right place
*    Add SETTINGS ACK Flag

Double-checking FTW.

Of course, there's a good chance I also messed something up here.  So feel free to check my work also.

Lukasa:
I think the language about ordering is a bit weird, but it is clarified somewhat by the hex representations shortly after. I think those hex representations are doing the heavy lifting, as I’m not aware of any issue with interoperability caused by getting the flag indices wrong in practice.

Nonetheless, if we’re really worried about it we could just delete “bit n” and replace it with “the x bit”, e.g. delete “bit 0” and replace it with “the `END_STREAM` bit”.

PR #861, opened 2021-06-03
Improve on fix for #815

gregw:
Improve on #846 that fixes for #815 by adding extra clarity:
 + The validation for uppercase characters is no longer listed separately
 + It is clearly stated that violations of the full HTTP ABNF field definition MAY be treated as *Malformed*

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -2587,13 +2587,13 @@
           <name>HTTP Fields</name>
           <t>
             HTTP fields carry information as a series of field lines, which are key-value pairs. For a listing of
-            registered HTTP fields, see the "Hypertext Transfer Protocol (HTTP) Field Name Registry" registry maintained at <eref target="https://www.iana.org/assignments/http-fields/"/>.
+            registered HTTP fields, see the "Hypertext Transfer Protocol (HTTP) Field Name Registry"
+            registry maintained at <eref target="https://www.iana.org/assignments/http-fields/"/>.
           </t>
           <t>
             Field names are strings of ASCII characters that are compared in a case-insensitive
             fashion. Field names MUST be converted to lowercase when constructing a HTTP/2
-            message. A request or response containing an uppercase character ('A' to 'Z', ASCII 0x41
-            to 0x5a) in a field name MUST be treated as <xref target="malformed">malformed</xref>.
+            message.
           </t>
           <t>
             HPACK is capable of carrying field names or values that are not valid in HTTP.  Though
@@ -2601,9 +2601,9 @@
           </t>
           <ul>
             <li>
-              A field name MUST NOT contain characters in the range 0x00-0x20 or 0x7F-0xFF (both
-              ranges inclusive).  This limits field names to visible ASCII characters, other than
-              ASCII SP (0x20).
+              A field name MUST NOT contain characters in the range 0x00-0x20, 0x41-0x5A or 0x7F-0xFF
+              (all ranges inclusive).  This limits field names to visible ASCII characters, other than
+              ASCII SP (0x20) and uppercase characters ('A' to 'Z', ASCII 0x41 to 0x5a).
             </li>
             <li>
               With the exception of <xref target="PseudoHeaderFields">pseudo-header fields</xref>,
@@ -2624,6 +2624,11 @@
             be treated as <xref target="malformed">malformed</xref>.  In particular, an intermediary
             that does not process fields when forwarding messages MUST NOT forward fields that
             contain any of the values that are listed as prohibited above.
+          </t>
+          <t>
+            In addition to the minimum validation defined above, non pseudo-header fields
+            that violate the HTTP ABNF grammar from <xref target="HTTP" section="5"/> MAY also be
+            treated as <xref target="malformed">malformed</xref>.
           </t>
           <t>
             Field values that are not valid according to the definition of the corresponding field
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -2587,13 +2587,13 @@
           <name>HTTP Fields</name>
           <t>
             HTTP fields carry information as a series of field lines, which are key-value pairs. For a listing of
-            registered HTTP fields, see the "Hypertext Transfer Protocol (HTTP) Field Name Registry"
-            registry maintained at <eref target="https://www.iana.org/assignments/http-fields/"/>.
+            registered HTTP fields, see the "Hypertext Transfer Protocol (HTTP) Field Name Registry" registry maintained at <eref target="https://www.iana.org/assignments/http-fields/"/>.
           </t>
           <t>
             Field names are strings of ASCII characters that are compared in a case-insensitive
             fashion. Field names MUST be converted to lowercase when constructing a HTTP/2
-            message.
+            message. A request or response containing an uppercase character ('A' to 'Z', ASCII 0x41
+            to 0x5a) in a field name MUST be treated as <xref target="malformed">malformed</xref>.
           </t>
           <t>
             HPACK is capable of carrying field names or values that are not valid in HTTP.  Though
@@ -2601,9 +2601,9 @@
           </t>
           <ul>
             <li>
-              A field name MUST NOT contain characters in the range 0x00-0x20, 0x41-0x5A or 0x7F-0xFF
-              (all ranges inclusive).  This limits field names to visible ASCII characters, other than
-              ASCII SP (0x20) and uppercase characters ('A' to 'Z', ASCII 0x41 to 0x5a).
+              A field name MUST NOT contain characters in the range 0x00-0x20 or 0x7F-0xFF (both
+              ranges inclusive).  This limits field names to visible ASCII characters, other than
+              ASCII SP (0x20).
             </li>
             <li>
               With the exception of <xref target="PseudoHeaderFields">pseudo-header fields</xref>,
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ venv/
 archive.json
 report.xml
 /lib
+/.idea/


gregw:
@martinthomson I think these are good simplifications/clarifications

martinthomson (approved):
Thanks @gregw.  I like this.  I don't think it changes the previous very much (and it improves the A-Z handling).  @Lukasa?

draft-ietf-httpbis-http2bis.xml
```diff
@@ -2587,23 +2587,23 @@
...
             HPACK is capable of carrying field names or values that are not valid in HTTP.  Though
             HPACK can carry any octet, fields are not valid in the following cases:
           </t>
           <ul>
             <li>
-              A field name MUST NOT contain characters in the range 0x00-0x20 or 0x7F-0xFF (both
-              ranges inclusive).  This limits field names to visible ASCII characters, other than
-              ASCII SP (0x20).
+              A field name MUST NOT contain characters in the range 0x00-0x20, 0x41-0x5A or 0x7F-0xFF
+              (all ranges inclusive).  This limits field names to visible ASCII characters, other than
```
mnot, line 2605:
```suggestion
              (all ranges inclusive).  This limits field names to visible ASCII characters, excluding
```

'other than' reads a little oddly to me...

draft-ietf-httpbis-http2bis.xml
```diff
@@ -2625,6 +2625,11 @@
             that does not process fields when forwarding messages MUST NOT forward fields that
             contain any of the values that are listed as prohibited above.
           </t>
+          <t>
+            In addition to the minimum validation defined above, non pseudo-header fields
+            that violate the HTTP ABNF grammar from <xref target="HTTP" section="5"/> MAY also be
+            treated as <xref target="malformed">malformed</xref>.
+          </t>
```
mnot, line 2632: This encourages sending a HTTP/2-specific error in situations where a generic HTTP semantic is more appropriate. 

For example, when a request header violates the ABNF and the recipient wants to reject the request, the appropriate thing to do is to send a 400 Bad Request status code, not close the connection with a HTTP/2 error.

Even for response headers (where there isn't an in-protocol mechanism in generic HTTP semantics), it's not clear that this is the right thing to do; treating the response as malformed means that the connection is closed, and intermediary senders often don't have full control over the syntax of headers they send -- and we shouldn't encourage them to attempt such control.

10.5 already [allows](https://httpwg.org/http2-spec/draft-ietf-httpbis-http2bis.html#name-denial-of-service-considera) endpoints to handle suspicious activity with a connection error. I think that covers the use case.

If we do end up saying something here, it needs to be squared with the considerations above, noting that HTTP [explicitly allows](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#error.handling) recipients to 'attempt to recover a usable protocol element from an invalid construct.'

mnot (changes requested):

draft-ietf-httpbis-http2bis.xml
martinthomson: I was OK with this on the basis that we previously permitted this behaviour, but that was a poorly considered requirement.  This is an opportunity to do better.

Would you prefer that we not say anything in this case?  It seems like we already allow for this possibility (in 10.5 as you say, but also in the core semantics docs), so maybe this is just unnecessary repetition.  And, as you say, unnecessary becomes harmful when the repetition isn't sufficiently precise.

A maybe-nit:

> treating the response as malformed means that the connection is closed

I don't know that this is right.  The requirement is to discard the response, so I assume that RST_STREAM is the expected reaction.

mnot: Ah, correct, I misremembered how malformed was treated. However, this MAY is seemingly in conflict with the requirements here:

> Malformed requests or responses that are detected MUST be treated as a stream error (Section 5.4.2) of type PROTOCOL_ERROR.
>
>... Clients MUST NOT accept a malformed response.

Also the [description of malformed](https://httpwg.org/http2-spec/draft-ietf-httpbis-http2bis.html#section-8.1.2.6) does not include HTTP header values.

Lukasa: I don't think this MAY is in conflict. This MAY grants the endpoint a new option when it observes a field that fails to validate according to the ABNF, which is to send a RST_STREAM. The MAY is "MAY treat as malformed": if you do, then the MUSTs above apply, but if you don't, they don't. 

However, I am sympathetic to the idea that failing to validate against the -semantics ABNF should not be a HTTP/2 protocol level error. I think this is happening to us because of two issues:

1. -semantics doesn't really clearly _say_ what you should do if you get a field that doesn't meet the ABNF requirements. 400 is implied by a careful read of the spec, but not required or even suggested. One part of -semantics talks about "senders MUST NOT emit" data that doesn't meet the ABNF, a different section quite far away says that "receivers MAY try to recover useful data" and then has nothing else to say. So I'm sympathetic to @gregw's desire to have a single place that outlines your options, especially as HTTP/2 is placing additional constraints.
2. Why do we have different reporting mechanisms for these? In HTTP/1 if you include a space in your field-name, you can expect a 400, while in HTTP/2 we mandate a RST_STREAM. If that mandate applied only to intermediaries then that would make some amount of sense, but it doesn't, it applies to all HTTP/2 implementations.

It is possible that we cannot reconcile this with -semantics: HTTP/2 is trying to apply strict rules that eliminate some bad cases that -semantics explicitly permits. These two things seem inherently in conflict to me.

Lukasa: Perhaps we should consider simply abandoning the change as impossible to reconcile with -semantics.

gregw: Prior to #846 the text in the intermediary section was:
> Any request or response that contains a character not permitted in a field value MUST be treated as <xref target="malformed">malformed</xref>.  Valid characters are defined by the <tt>field-content</tt> ABNF rule in <xref target="HTTP" section="5.5"/>

This has now been modified to: 
> An intermediary can reject fields that contain invalid field names or values for other  reasons, in particular those that do not conform to the HTTP ABNF grammar from <xref target="HTTP" section="5"/>.

So current implementations have been allowed to treat non HTTP compliant as *malformed*  if they were an intermediary. Now we are just saying "reject fields", but not defining how that is done.

I like the clarity that after #846, a field-name with space in it MUST be treated as malformed.
Yet a field-name with a doublequote will be treated as malformed if handled by an intermediary under the previous version; or rejected in and undefined way by an intermediary in the current version; or perhaps handled with a 400 bad request if not an intermediary.  That is just totally confusing.

Either sending an invalid HTTP field-name is a h2 protocol error (reset-stream) or it is a http semantic error (400 bad request).    

I think that if the field violates the allowed alphabet for HTTP valid fields, then it is a h2 protocol error: reset-stream.
But if the field is invalide because of semantics ( multiple fields, missing fields, bad field values) then it is a http protocol error: 400 bad request

Sorry - to relitigating #846,  but I'm unsure why it has created a small set of fields that are inbetween the two of those - ie fields with double quotes in their field name.  Are there any sensible fields that are invalid http but valid h2#846 that are actually are needed by anything?

Eitherway, this document has to define what can be done with such fields (valid h2 fields by #846 but not valid http2 fields).    If they are treated as h2 errors, then there is no difference to just excluding those few extra characters from the minimum validation in the first place.   If are treated as http errors, then that is double handling as field names and values will need to be scanned twice to validate the alphabet used. Insert rant about carbon foot print here.

mnot: Focusing on a specific case is helpful, thanks.

Just to get it out of the way -- In a perfect world we'd specify exactly one way to do this and all implementations would align on that; doing so would almost certainly improve security. Unfortunately, we don't live in that world, because there are already a large number of deployed implementations with diverse behaviour that may not change, and because some implementations don't want to touch every byte flowing through them.

That said, the recipient of an invalid HTTP field-name has a choice:
 
1. If they can recover from it, they can ignore the invalid bit and pass it on to the consuming application or next hop (e.g., a `"` is recoverable; a `\n` is not)
2. If it's a request, they can respond with a 400 (regardless of whether it's an intermediary)
3. If h2 or h3 is in use, they can RST_STREAM with an error code (because it's considered malformed above, I think this would be `ENHANCE_YOUR_CALM`

I'd argue that the ordering above is a first-match-wins approach that most implementations should use, although an implementation can choose to skip (1), and we might want to highlight the security considerations therein -- but we can't make that an across-the-board requirement.

field-values, on the other hand, are much broader; [their definition](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#fields.values) only excludes most control characters (due to `obs-text`). That said, the same approach can be used above for control characters.

Regarding how it's documented -- I'd very much prefer that prose rules, NOT references to ABNF, be used here, because talking about field-value ABNF is going to confuse some readers into thinking that they need to reject headers based upon their _individual_ ABNF (at first that's what I thought you were proposing).

gregw: The key thing I'm after is the clarity that an implementation MAY exceed the minimum field syntax validation and treat such messages as malformed. Ie it needs to be clear that the minimum is just the minimum field syntax validation and that it MAY be exceeded by fully checking HTTP field syntax.

I want to be able to simply implement a single test for field syntax that will equally handle a space or double quote, without the risk that some user then comes and complains that I'm over validating and that they should be able to send double quotes in field names.

Nor do i want to have to check minimum h2 field syntax, then HTTP field syntax separately. Some implementations might want to do minimal validation and then recover from a field with a double quote, but I want it to be ok not to do that and to just reset steam



The reference to HTTP ABNF is exactly the same as the intermediary section later in the document. If there is another way to say it without ABNF, then fine, but that would be the third definition of field syntax validity in the document.

gregw: Also, if this text is not ok, that says an impl MAY NOT validate against http ABNF , which is in contradiction to the section on intermediaries?

mnot: Greg, I think we're largely in agreement -- my text above is arguing for adding more nuance / information, not fundamentally against what you're proposing.

The one place where I *potentially* see disagreement is with 'I want it to be ok not to do that and to just reset st[r]eam'. It would be really good if we had a common understanding as to whether this is the right thing to do when a server receives a request and has the opportunity to generate a 400. I think that needs wider discussion.

gregw: @mnot  I think the split should be syntax versus semantics.  If a field breaks the field syntax (ie grammar) for h2 or HTTP, then it is a protocol error as it is a field should never have been sent.
Whilst if the field breaks HTTP semantics (eg. wrong type for the value, duplicate etc.) then it is a field that can be sent but that can't be understood, so it should be handled with a  400 (or perhaps ignorable or rejected in some other way).

So the nuance that you are after is that for fields in the syntax validation gap between minimal h2 and HTTP you want to give implementations some flexibility to treat them like semantic issues.  So perhaps text like:

```suggestion
          <t>
            In addition to the minimum validation defined above, non pseudo-header fields
            that violate the HTTP ABNF grammar from <xref target="HTTP" section="5"/> MAY either: be
            treated as <xref target="malformed">malformed</xref>; be ignored; or if part of a request then responded to 
            with a 400 Bad Request response.
          </t>
```

But then do we need to say that 400 responses and corrections are also applicable to fields that violate the HTTP semantics?  or is that just assumed?

Plus also, for my own implementation, I really would not want to allow field-names with double quotes into the phase of being semantically evaluated.  It is probably safe to do so, but that code was written and tested against the HTTP semantics which do not allow double-quotes, so I really have no idea that it is safe to let such field propagate into it.  It's also somewhat user extensible and I have even less idea what downstream problems could happen if I let such fields through.  Nor do I want to enhance that semantic layer to do another round of syntax validation and carbon emissions generation.

Oh no... I find myself yet again relitigating... if we just said the h2 implementations MUST NOT send fields that break the HTTP syntax, then there would be no gap and thus no problem.  I keep coming back to that because I'm not understanding the use-case for allowing a message with a field-name containing a double quote?  If we just took a few more characters out of the acceptable alphabet for h2 field validation, then this issue goes away and I don't think we break anybody?

draft-ietf-httpbis-http2bis.xml
```diff
@@ -2587,23 +2587,23 @@
...
             HPACK can carry any octet, fields are not valid in the following cases:
           </t>
           <ul>
             <li>
-              A field name MUST NOT contain characters in the range 0x00-0x20 or 0x7F-0xFF (both
-              ranges inclusive).  This limits field names to visible ASCII characters, other than
-              ASCII SP (0x20).
+              A field name MUST NOT contain characters in the range 0x00-0x20, 0x41-0x5A or 0x7F-0xFF
+              (all ranges inclusive).  This limits field names to visible ASCII characters, excluding
+              ASCII SP (0x20) and uppercase characters ('A' to 'Z', ASCII 0x41 to 0x5a).
```
gregw, line 2606: @mnot to get rid of the gap problem, we could just increase the minimal validation to exclude all field names that 
break HTTP syntax (but not semantics).
```suggestion
              A field name MUST NOT contain characters 0x00-0x20, 0x22, 0x28, 0x29,
              0x2C, 0x41-0x5D, 0x7B, 0x7D or 0x7F-0xFF (all ranges inclusive).
              This limits field names to visible ASCII characters permitted by the HTTP ABNF 
              grammar from <xref target="HTTP" section="5"/>, other than uppercase characters 
              ('A' to 'Z', ASCII 0x41 to 0x5a).
```
The down side ofthis is a slightly more difficult check expression, that probably needs to dereference to a table.

gregw:
@martinthomson should I spin out the non controversial change (of combining the lowercase validation into the minimal validation bullet point) into it's own PR so that it doesn't get lost if this PR never reaches a conclusion?

martinthomson:
If you could, that might help.  I want to talk about the tricky question at our interim next week, but that other piece I can just merge.

gregw:
The non controversial changes have been moved to #864

draft-ietf-httpbis-http2bis.xml
gregw: I'll resolve this comment, because even if it is agreed to validate against the HTTP field name syntax, this is a horrid way to specify it!

draft-ietf-httpbis-http2bis.xml
gregw: To give some more concrete data, here is [Jetty's implementation of h2 field-name validation](https://github.com/eclipse/jetty.project/blob/jetty-11.0.4/jetty-http2/http2-hpack/src/main/java/org/eclipse/jetty/http2/hpack/HpackDecoder.java#L170-L198).  You can see that it treats any violations of either h2 or http field criteria as a streamException (ie reset).

At the very least, it would be really good if this update of the specification did not make this existing (and used) implementation non compliant and needing to be rewritten to in order to send 400 responses when used in a server.

It would be great if the specification explicitly said that a compliant implementation MAY be implemented like this, so we don't have to deal with users trying to send bizarre field-names and complaining we are not compliant with the spec.

gregw:
@martinthomson any feedback from the interim meeting about how I can make this PR more palatable?

draft-ietf-httpbis-http2bis.xml
gregw: @mnot, I really think something does need to be said here, since as currently worded it makes the jetty implementation non-compliant, in that we treat fields that break the ABNF as malformed, but send 400s for any semantic issues.    This was allowed in the previous versions, but now this document says "fields are not valid in the following cases:"  without any caviate to indicate that it is minimal validation (eg  "fields are not valid in at least the following cases:")

The intention was to require minimal validation, but as written the change is specifying exact validation since a reasonable reading of this section would suggest that any fields that are not-not-valid are valid.   So to enforce implementation that were not validating, you have now made jetty, which has always validated, non compliant!

To resolve this against the document as written, I would need to do a second round of character checking of the field above the hpack layer to issue a 400.   It will be difficult for me to do this without impossing this second character checking on all protocols: HTTP/1, FCGI, etc. as by the time I'm able to send 400 responses I am in protocol neutral code.    This will be a significant impact with a huge carbon footprint as my server deals with some very high volume websites.

mnot:
@gregw I took an action to work on a PR - see https://github.com/httpwg/http2-spec/pull/866

gregw:
Closed in favour of #866

PR #864, opened 2021-06-10
Improve on fix for #815 for lowercase validation

gregw:
The validation for uppercase characters is no longer listed separately, but instead included with the minimal validation.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -2601,7 +2601,7 @@
           </t>
           <ul>
             <li>
-              A field name MUST NOT contain characters in the ranges 0x00-0x20, 0x41-0x5A or 0x7F-0xFF
+              A field name MUST NOT contain characters in the ranges 0x00-0x20, 0x41-0x5A, or 0x7F-0xFF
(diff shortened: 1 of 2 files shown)


gregw:
@martinthomson : factored out non controversial changes from #861

draft-ietf-httpbis-http2bis.xml
```diff
@@ -2587,23 +2587,23 @@
...
           <t>
             HPACK is capable of carrying field names or values that are not valid in HTTP.  Though
             HPACK can carry any octet, fields are not valid in the following cases:
           </t>
           <ul>
             <li>
-              A field name MUST NOT contain characters in the range 0x00-0x20 or 0x7F-0xFF (both
-              ranges inclusive).  This limits field names to visible ASCII characters, other than
-              ASCII SP (0x20).
+              A field name MUST NOT contain characters in the ranges 0x00-0x20, 0x41-0x5A or 0x7F-0xFF
```
martinthomson, line 2604:
```suggestion
              A field name MUST NOT contain characters in the ranges 0x00-0x20, 0x41-0x5A, or 0x7F-0xFF
```

martinthomson (approved):

PR #866, opened 2021-06-16
Discuss invalid field names and values

### LINKED ISSUE #865: Errors at the border of framing and semantics
Validating fields for characters has raised a concern where we don't have a clear delineation between framing and semantics.  So the way that endpoints handle those cases is unclear.  For requests in particular, the choice between resetting a stream and sending a 4xx response is unclear.

It would be good if we could have more consistent overall behaviour here, but we have seen from the character composition issue (see #815) that different implementations will enforce the rules at different layers.  We probably need to allow for some flexibility.

@mnot has volunteered to help us navigate this somewhat nuanced issue.

mnot:
Fixes #865

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -2626,18 +2626,21 @@
             contain any of the values that are listed as prohibited above.
           </t>
           <t>
-            A recipient MAY perform additional validation on field names and values, to exclude
+            A recipient MAY treat a message that contains a field name or value that includes
             other characters disallowed by <xref target="HTTP" section="5.1"/> and <xref
-            target="HTTP" section="5.5"/>, respectively. When such a recipient wishes to reject
-            an invalid request on this basis, it SHOULD do so using the 400 (Bad Request) status code
-            <xref target="HTTP" section="15.5.1"/>. If the message is a response, or the status
-            code cannot be sent (e.g., because the error occurs in a trailer field), the recipient MAY
-            consider the message to be <xref target="malformed" format="none">malformed</xref>.
+            target="HTTP" section="5.5"/> as <xref target="malformed">malformed</xref>.
           </t>
           <t>
-            Field values that are not valid according to the definition of the corresponding field
-            do not cause a request to be <xref target="malformed" format="none">malformed</xref>
-            except as defined by the processing rules for the field.
+            When a request message violates one of the requirements above, it SHOULD be responded
+            to using the 400 (Bad Request) status code <xref target="HTTP" section="15.5.1"/>
+            before the stream is reset, unless a more suitable status code is defined, or the
+            status code cannot be sent (e.g., because the error occurs in a trailer field).
+          </t>
+          <t>
+            Note that field values that are not valid according to the definition of the corresponding field
+            do not cause a request to be <xref target="malformed" format="none">malformed</xref>; the
+            requirements above only apply to the generic syntax for field values as defined in <xref
+            target="HTTP" section="5.5"/>.
           </t>
           <section anchor="PseudoHeaderFields">
             <name>Pseudo-Header Fields</name>
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -2626,6 +2626,15 @@
             contain any of the values that are listed as prohibited above.
           </t>
           <t>
+            A recipient MAY perform additional validation on field names and values, to exclude
+            other characters disallowed by <xref target="HTTP" section="5.1"/> and <xref
+            target="HTTP" section="5.5"/>, respectively. When such a recipient wishes to reject
+            an invalid request on this basis, it SHOULD do so using the 400 (Bad Request) status code
+            <xref target="HTTP" section="15.5.1"/>. If the message is a response, or the status
+            code cannot be sent (e.g., because the error occurs in a trailer field), the recipient MAY
+            consider the message to be <xref target="malformed" format="none">malformed</xref>.
+          </t>
+          <t>
             Field values that are not valid according to the definition of the corresponding field
             do not cause a request to be <xref target="malformed" format="none">malformed</xref>
             except as defined by the processing rules for the field.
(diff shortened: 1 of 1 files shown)


Lukasa (approved):
I think this is reasonable.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -2625,6 +2625,15 @@
             that does not process fields when forwarding messages MUST NOT forward fields that
             contain any of the values that are listed as prohibited above.
           </t>
+          <t>
+            A recipient MAY perform additional validation on field names and values, to exclude
+            other characters disallowed by <xref target="HTTP" section="5.1"/> and <xref
+            target="HTTP" section="5.5"/>, respectively. When such a recipient wishes to reject
+            an invalid request on this basis, it SHOULD do so using the 400 (Bad Request) status code
+            <xref target="HTTP" section="15.5.1"/>. If the message is a response, or the status
+            code cannot be sent (e.g., because the error occurs in a trailer field), the recipient MAY
```
martinthomson, line 2634: I just want to poke on this "MAY" here a little.  I think it's right, but I want to test it.

This is "MAY" on the basis that recipients might not detect an error, not because - when an error is detected - we want the recipient to mostly ignore that error, right?

What I'm trying to get at is whether the structure here is:

MAY check
if checked then:
* SHOULD send 400 when appropriate
* otherwise, treat as malformed

In which case the otherwise could be a MUST.  Though the wording chosen here is such that it would make a poor MUST, I think that I might prefer something stronger.

The other factor arguing against a stronger statement is that the SHOULD allows for requests to be treated with other than a 400 response.  That is, it effectively allows errors to be ignored or RST_STREAM to be sent instead.  We want that, but it makes it hard to say "SHOULD ... unless".  I also don't want to touch the existing SHOULD statement, which is good in its current form.

My preference then might be to say SHOULD here rather than MAY.

```suggestion
            code cannot be sent (e.g., because the error occurs in a trailer field), the recipient SHOULD
```

martinthomson (commented):
I like this, but I want to make sure that we have captured the desired logic.

draft-ietf-httpbis-http2bis.xml
mnot: Your analysis is correct, but I think it would be better to convert MAY into a 'can', so that readers aren't confused about the (not terribly obvious) link between it and the preceding MAY. Does that work for you?

gregw: Firstly this is a big improvement on the current text, as it no longer makes existing good implementations non-compliant.

However, I still do take issue with "SHOULD do so using the 400" as I believe that you have applied that too broadly.

I agree that semantic errors in fields (duplicate fields, missing fields, wrong value types etc.) SHOULD be responded to with a 400, but I do not think that applies for basic violations of field grammer as defined by the ABNF (ie the example of a field name including a double quote) as defined by sections 5.1 and 5.5 of the referenced HTTP document.

It is confusing and over complex to require that a field containing `:` is malformed, but one containing `"` is h2 is a bad-request sometimes and malformed othertimes.  

It is not efficient to require two validations for field grammer: once for the h2 requirements (malformed if it contains a ':' or 'A') and then again for HTTP (400 if it contains a double quote).  This will be a massive carbon foot print as every single field will need to be scanned a second time (or a more complex single scan) just to check for the vanishly small possibility that there is a double quote. 

I think that recipients MAY validate against 5.1 & 5.5 and when they do they should reject fields as malformed.  Recipients MAY perform additional validation against the semantics of HTTP fields (ref whole document), in which case they SHOULD reject requests with 400.

gregw (changes requested):

draft-ietf-httpbis-http2bis.xml
martinthomson: This seems to be about two things : division of responsibilities (400 is core semantics, RST_STREAM is h2) and where enforcement responsibilities lie as a result.  I think that we've learned that the character composition of fields (names or values) is split, but enforcement might occur at either layer.

What Mark's text attempts to do (and I think it succeeds) is say that when the rule is in core semantics, then the enforcement should look like a semantic-layer rejection.  That's a principled position.  

However, this acknowledges the pragmatic position and allows it (it's not MUST): if you detect ":" and send RST_STREAM, that's OK too.  It just recommends (with SHOULD) that you try to engage semantic-layer signals, because those are better.  On principle, I agree with this view.

That some implementations combine these checks and use what amount to less rigorous signals, doesn't make them bad.  It is just a trade-off they make: efficient processing over clearer signaling.  As this is a serious error that will likely engage engineers and not users, what signal is used doesn't matter THAT much.

(As for the "MAY" vs. "can", "can" works for me.)

gregw: @martinthomson but I don't think the text does achieve allowing for semantic rejection of  invalid fields when possible.  It still leaves the MUST be treated as malformed for any field with a `:` or `A`, etc.  

If sending a 400 when possible is a principled position, then the principle should apply to all invalid fields and  I don't see why the fields that are at risk of creating a security problem (and this invalidated at the h2 level) are exempted from the principle?  The text should say that when possible a receiver SHOULD reject a malformed field with a 400 regardless of the reason for the malformation.  

This proposed text is that it still encourages two classes of invalid fields with different behaviours for those classes.  Thus implementors will either simply follow the document and be less efficient or be more efficient and considered "less rigorous".

mnot: Yes - that's why it's a SHOULD. And conforming with that SHOULD will not require scanning a second time for every request -- only when there's a need to distinguish between the two errors. 

The reason that promoting this to the semantic layer when possible is important is that it's more likely to reach that engineer, through protocol hops and APIs that don't know about h2 error codes.

gregw: @mnot, what I'm not understanding is why you see two types of error?   I'm not disputing how they should be handled, only that the spec should allow all syntax validation failures of fields to be handled equally.   Ie SHOULD send a 400 for a field name with a space just as much as for one with a double quote.   If you can't send a 400, then treat as malformed.

The validation was added to the spec for security reason, because space, `:` and control characters are potentially dangerous if forwarded.  But I just don't see that makes them a different class of error to a field with a double quote and no less deserving of a clear signal that will reach an engineer.   If anything they are more deserving of a clear signal as they very likely indicate an attack.

So can you explain to me why you see a "need to distinguish between" an invalid field with a space vs one with a double quote?  Why can't  the spec be written in a way to encourage best practise for both?

gregw: I think we are making heavy work of this section by trying to treat a 400 response as something different to malformed message handling.  Malformed message handling already allows a response code to be sent at MAY strength. I think the text is simpler if we just say treat as malformed and then add a SHOULD for 400 responses to requests if possible.

The additional validation can then be simply handled as an optional bullet point in the list of validations.  I have created PR #868 with these changes.

gregw:
See #868 as an alternative.

draft-ietf-httpbis-http2bis.xml
mnot: @gregw I think you're right; pushing a revision shortly.

gregw (approved):
LGTM

martinthomson:
Thanks everyone for your patience here.  It's good to see this work out :)

PR #868, opened 2021-07-07
Alternate for #866 Invalid field handling

gregw:
Removed distinction between treating as malformed and sending a 400, since treating as malformed already
allows a 400 to be sent.

This uses some text from #866 and replaces that PR.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -2618,12 +2618,20 @@
               A field value MUST NOT start or end with an ASCII whitespace character (ASCII SP or
               HTAB, 0x20 or 0x9).
             </li>
+            <li>
+              Optionally, a field name or field value MAY NOT contain characters disallowed
+              by <xref target="HTTP" section="5.1"/> and <xref target="HTTP" section="5.5"/>,
+              respectively.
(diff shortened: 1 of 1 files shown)


gregw:
The outcome I'm striving for here is to avoid the situation where the spec has two classes of invalid fields.  With #866 it currently has that a message with one type of invalid character MUSTS be treated as malformed and that bad response code MAY be sent;   but for some other types of invalid characters it just says a bad response code SHOULD be sent otherwise MUST treated as malformed  (which then says a bad response code MAY be sent).

This PR simplies it to saying that all messages with invalid characters MUST be treated as  malformed and that if the message is a request then a bad response code SHOULD be sent.    
The definition of invalid characters optionally include those disallows by HTTP.

gregw:
@mnot @martinthomson

martinthomson:
So this makes the recommendation to use 400 stronger; extending it to include more fields.  I think that's OK, but I'm going to defer to @mnot on this.  I'm OK with either.

mnot:
I did a rewrite, PTAL. Happy to review this one but I'd be suggesting a number of edits here.

gregw:
closed in favour of the updated #866

PR #871, opened 2021-07-07
Change log for character validation

### LINKED ISSUE #862: Revise acknowledgments
I don't know what the convention is here, but the acknowledgments seem a little off.  For instance, we acknowledge people for mechanisms that were removed in this revision.

We might be able to remove these entirely; we could simplify (which would avoid problems like the above); or we might just mark these as applying to RFC 7540.  Ideas sought.

martinthomson:
The only change we have planned for the draft that isn't in the change
log yet.

Captures #865 changes, whether that is via #866 or other changes, the
substance of the change is roughly the same.

Closes #862.


martinthomson:
Oops, closes the wrong issue.  I'll try again.

PR #873, opened 2021-07-12
A simple, but manual, s/bit \d/the XXX Flag/

martinthomson:
Though I had to be careful with text on TCP, which has FIN and RST bits
we don't want to touch.


Lukasa (approved):
LGTM.

PR #874, opened 2021-07-12
HTTP work

mnot:
This is a rearrangement of the HTTP-related sections, in preparation for actually addressing 867.


mnot:
This doesn't actually address the issue -- I came to a point where I realised this would add a bunch of new requirements and prose that would need heavy scrutiny, and if we want to finish soon, that may not be advisable (although maybe we can figure it out during WGLC). 

In the meantime, I think this rearrangement is *much* clearer than the original ordering -- see what you think.

PR #877, opened 2021-07-23
Add some structure to the undifferentiated mess

martinthomson:
...that is frame formats.

I've chosen to put blank lines before and after both frames and the
stream identifier part of the stream header.  That should make this a
lot clearer.

@wtarreau was right that this was a little hard to process without that.
Hopefully this is a tiny bit better.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -419,9 +419,12 @@
 HTTP Frame {
   Length (24),
   Type (8),
+
   Flags (8),
+
   Reserved (1),
   Stream Identifier (31),
+
   Frame Payload (..),
 }
 ]]></artwork>
@@ -1331,12 +1334,15 @@ HTTP Frame {
 DATA Frame {
   Length (24),
   Type (8) = 0,
+
   Unused Flags (4),
   PADDED Flag (1),
   Unused Flags (2),
   END_STREAM Flag (1),
+
   Reserved (1),
   Stream Identifier (31),
+
   [Pad Length (8)],
   Data (..),
   Padding (..),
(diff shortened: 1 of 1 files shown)


mnot:
I think this helps. The only other clarification I can think of is to pull the flags out into a separate illustration, but let's see if this is sufficient first.

wtarreau:
Bah I didn't respond at the right place. One day I'll understand github's interface. Or not :-)

Both changes look good and possibly sufficient to me. Let's just wait a bit to see if anyone in the WG has any better proposal. Thanks for your quick reaction!

wtarreau:
For the flags (and only the flags), what about trying this:
```
DATA Frame {
  Length (24),
  Type (8) = 0,

  Unused Flags (4),    // 0xF0
  PADDED Flag (1),     // 0x08
  Unused Flags (2),    // 0x06
  END_STREAM Flag (1), // 0x01

  Reserved (1),
  Stream Identifier (31),
  [Pad Length (8)],
  Data (..),
  Padding (..),
```

Lukasa (approved):

martinthomson:
Should I add the octet values of flags?  I would only do that for non-unused flags though.

Lukasa:
I believe the prose text contains them, so there's at least a defensible reason for leaving them off.

wtarreau:
They're indeed present but not trivially spottable. I have a dead-tree copy of 7540 here with them overlined in pink to spot them faster :-/  We could save a tree by noting them there :-). I agree with Martin that we can avoid the unused ones though, because what matters in tests is not to know the unused ones but rather to figure them as the complement of all known ones. I won't roll over the floor crying if that's rejected, but as an implementer I find that it significantly helps during debugging.

martinthomson:
As I don't know *how* to comment in this format and I don't want to start inventing something (else), I think I'm going to take the coward's option.

mnot:
What are the semantics of a blank line in this format?

/me ducks

PR #879, opened 2021-08-02
Clarify the rules on server-sent HEADERS

### LINKED ISSUE #878: Unpromised Push
As a corollary to https://github.com/quicwg/base-drafts/issues/4930, I cannot find anything in RFC 7540 that explicitly prohibits the server sending a response on a stream that hasn't been promised.  The closest is the state machine, where a push stream is supposed to go from idle to reserved to half-closed.  The statement that it can go from idle to open upon sending a HEADERS frame is clearly oriented at the client, but is endpoint-agnostic:  `Sending or receiving a HEADERS frame causes the stream to become "open".`

Lukasa:
Resolves #878.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -704,8 +704,9 @@ HTTP Frame {
             </t>
             <ul spacing="normal">
               <li>
-                    Sending or receiving a <xref target="HEADERS" format="none">HEADERS</xref> frame causes the stream to become
-                    "open".  The stream identifier is selected as described in <xref target="StreamIdentifiers"/>.  The same <xref target="HEADERS" format="none">HEADERS</xref> frame can also
+                    Sending a <xref target="HEADERS" format="none">HEADERS</xref> frame as a client, or receiving a HEADERS frame
+                    as a server, causes the stream to become "open".  The stream identifier is selected as described in
+                    <xref target="StreamIdentifiers"/>.  The same <xref target="HEADERS" format="none">HEADERS</xref> frame can also
                     cause a stream to immediately become "half-closed".
                   </li>
               <li>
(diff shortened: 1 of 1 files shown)


mnot:
Probably needs to be listed in "Changes from RFC7540".

Lukasa:
Does it? I think this change is editorial: this is already _effectively_ the rule, we're just calling it out explicitly.

mnot:
If by "effectively" you mean "it's what implementations do", we still need to document a change in the *rfc*.

martinthomson:
I read that as "it's what the document originally said, just not very clearly".

Lukasa:
Yup, that's my assertion, see the discussion in #878.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -727,6 +728,9 @@ HTTP Frame {
             <t>
                 Receiving any frame other than <xref target="HEADERS" format="none">HEADERS</xref> or <xref target="PRIORITY" format="none">PRIORITY</xref> on
                 a stream in this state MUST be treated as a <xref target="ConnectionErrorHandler">connection error</xref> of type
+                <xref target="PROTOCOL_ERROR" format="none">PROTOCOL_ERROR</xref>. If this stream is server-initiated as described in
+                <xref target="StreamIdentifiers"/> then receiving a <xref target="HEADERS" format="none">HEADERS</xref> frame MUST also
+                be treated as a <xref target="ConnectionErrorHandler">connection error</xref> of type
```
MikeBishop, line 733: Feels like it could use at least one comma in all that.

MikeBishop (approved):
Yes, I think this is correct.

draft-ietf-httpbis-http2bis.xml
martinthomson:
```suggestion
                <xref target="PROTOCOL_ERROR" format="none">PROTOCOL_ERROR</xref>. If this stream is server-initiated, as described in
                <xref target="StreamIdentifiers"/>, then receiving a <xref target="HEADERS" format="none">HEADERS</xref> frame MUST also
                be treated as a <xref target="ConnectionErrorHandler">connection error</xref> of type
```
?

Issue #883, opened 2021-08-06
Prohibit multiple :authority pseudo-headers


nharper:
In the context of https://lists.w3.org/Archives/Public/ietf-http-wg/2021JulSep/0149.html, I noticed that there's no prohibition on a request having multiple `:authority` pseudo-headers. Such a request should be treated as malformed.

martinthomson:
Closed by #884.

PR #884, opened 2021-08-06
Prohibit duplicate pseudo-headers.

nharper:
This fixes issue #883.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -2829,6 +2829,9 @@ cookie: e=f
           All pseudo-header fields MUST appear in a field block before all regular field lines.
           Any request or response that contains a pseudo-header field that appears in a field
           block after a regular field line MUST be treated as <xref target="malformed">malformed</xref>.
+          The same pseudo-header field name MUST NOT appear more than once in a field block. A
+          field block for an HTTP request or response that contains a repeated pseudo-header field
+          name MUST be treated as <xref target="malformed">malformed</xref>.
         </t>
         <section anchor="HttpRequest">
           <name>Request Pseudo-Header Fields</name>
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -2829,6 +2829,8 @@ cookie: e=f
           All pseudo-header fields MUST appear in a field block before all regular field lines.
           Any request or response that contains a pseudo-header field that appears in a field
           block after a regular field line MUST be treated as <xref target="malformed">malformed</xref>.
+        </t>
+        <t>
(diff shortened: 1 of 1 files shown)


Lukasa (commented):
This seems like a reasonable addition, but I think it’s in the wrong place, for two reasons:

1. There’s no particular reason to exclude response pseudo-headers from this prohibition: multiple `:status` fields is just as meaningless.
2. We probably don’t want to prohibit this _generally_, but instead only for the pseudo-headers we define in this document.

To that end I think I recommend moving this up to the parent section, and noting something like “All pseudo-header fields defined in this document MUST NOT appear more than once in a header header section.”

nharper:
I agree on 1 that there's no reason to exclude response pseudo-headers from this requirement. (I was worried about contradicting something regarding multiple interim responses, each of which needs a `:status` field, but each of those are in a separate field section, so a rule of no duplicates within a field section should work for pseudo-headers, both request and response.)

However, I think that a general prohibition is a good thing. If a future (negotiated) extension wants to introduce a new pseudo-header that can be duplicated, it could override an otherwise general prohibition in this draft. If a future extension doesn't do that, it gets the more sane/safe default for free.

Lukasa:
Hmm, yeah, I think I can see a justification for that argument. Given that adding new pseudo-headers needs to be negotiated _anyway_, I don't think that's a big deal. So I'm happy to go ahead with the change for (1) and skip (2).

martinthomson (approved):
I had something else.  This is better.

nharper:
@martinthomson is "this" the current text, or revised text to address Cory's comment?

martinthomson:
This pull request.  I think that a blanket "at most once" is fine and what you have is enough.  Future extensions can deal with the consequences of that if they want >1.

Lukasa:
@martinthomson Do you want the change to cover responses as well?

martinthomson:
That would mean moving it up to the "HTTP Control Data" section.  That is a good idea.  I missed that this was under requests only.

nharper:
I moved the language up to the HTTP Control Data section and slightly tweaked it.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -2829,6 +2829,9 @@ cookie: e=f
           All pseudo-header fields MUST appear in a field block before all regular field lines.
           Any request or response that contains a pseudo-header field that appears in a field
           block after a regular field line MUST be treated as <xref target="malformed">malformed</xref>.
+          The same pseudo-header field name MUST NOT appear more than once in a field block. A
```
martinthomson, line 2832:
```suggestion
        </t>
        <t>
          The same pseudo-header field name MUST NOT appear more than once in a field block. A
```

Let's use paragraphs.

martinthomson (approved):

Lukasa (approved):

PR #894, opened 2021-08-13
Avoid h2c awkwardness

### LINKED ISSUE #892: nit: is h2c spec'd or not
Maybe a hangover from trying to tear up the photos of the "http" URI and upgrade dance. But this seems to read oddly. 

On one hand:

> This usage was never widely deployed, and is no longer specified in this document.

on the other:

> Negotiating "h2" or "h2c" implies the use of the transport, security, framing, and message semantics described in this document.

I don't have a solid suggestion to improve things.

LPardue:
fixes #892﻿

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -277,7 +277,9 @@
       <section anchor="versioning">
         <name>HTTP/2 Version Identification</name>
         <t>
-          The protocol defined in this document has two identifiers.
+          The protocol defined in this document has two identifiers. Creating a
+          connection based on either implies the use of the transport, security,
+          framing, and message semantics described in this document.
         </t>
         <ul spacing="normal">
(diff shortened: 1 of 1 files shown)


LPardue:
By removing the word negotiating and avoiding the actual identifiers, we can get the important point across

draft-ietf-httpbis-http2bis.xml
```diff
@@ -277,7 +277,9 @@
       <section anchor="versioning">
         <name>HTTP/2 Version Identification</name>
         <t>
-          The protocol defined in this document has two identifiers.
+          The protocol defined in this document has two identifiers. Creating a
+          connection based on either implies the use of the transport, security,
```
martinthomson, line 281: I realize now that "security" does not hold for both.  We might need to explicitly say that "h2c" does not provide the security properties of "h2".

PR #899, opened 2021-08-13
Make introduction more objective

### LINKED ISSUE #890: The reports of HTTP/1.1's negative performance are possibly exaggerated
...  and HTTP/2's positives too!

>    The Hypertext Transfer Protocol (HTTP) is a wildly successful
   protocol.  However, the way HTTP/1.1 uses the underlying transport
   ([RFC7230], Section 6) has several characteristics that have a
   negative overall effect on application performance today.

 >  In particular, HTTP/1.0 allowed only one request to be outstanding at
   a time on a given TCP connection.  HTTP/1.1 added request pipelining,
   but this only partially addressed request concurrency and still
   suffers from head-of-line blocking.  Therefore, HTTP/1.0 and HTTP/1.1
   clients that need to make many requests use multiple connections to a
   server in order to achieve concurrency and thereby reduce latency.

I think time has shown these claims to not be great. How about cutting cruft and saying

```
The performance of application using the Hypertext Transfer Protocol (HTTP) is
linked to how each version of HTTP uses the underlying transport, and the
conditions under which the transport operates.

In particular, HTTP message concurrency can reduce latency and improve
application performance. HTTP/1.0 allowed only one request to be outstanding at
a time on a given TCP connection. HTTP/1.1 added request pipelining, but this
only partially addressed request concurrency and still suffers from
application-layer head-of-line blocking. Therefore, HTTP/1.0 and HTTP/1.1
clients that need to make many requests use multiple connections to a server in
order to achieve concurrency.
```

and then a tweak to a later paragraph in the intro to state that no one is perfect

```
The resulting protocol is more friendly to the network because fewer TCP
connections can be used in comparison to HTTP/1.x. This means less competition
with other flows and longer-lived connections, which in turn lead to better
utilization of available network capacity. Note, however, that TCP head-of-line
blocking cannot be addressed by this protocol.
```

LPardue:
fixes #890﻿

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -67,16 +67,19 @@
     <section anchor="intro">
       <name>Introduction</name>
       <t>
-        The Hypertext Transfer Protocol (HTTP) is a wildly successful protocol. However, the way
-        HTTP/1.1 uses the underlying transport (<xref target="HTTP11"/>) has several
-        characteristics that have a negative overall effect on application performance today.
+        The performance of application using the Hypertext Transfer Protocol
+        (HTTP) is linked to how each version of HTTP uses the underlying
+        transport, and the conditions under which the transport operates.
       </t>
       <t>
-        In particular, HTTP/1.0 allowed only one request to be outstanding at a time on a given TCP
-        connection. HTTP/1.1 added request pipelining, but this only partially addressed request
-        concurrency and still suffers from head-of-line blocking. Therefore, HTTP/1.0 and HTTP/1.1
-        clients that need to make many requests use multiple connections to a server in order to
-        achieve concurrency and thereby reduce latency.
+        In particular, HTTP message concurrency can reduce latency and improve
+        application performance. HTTP/1.0 allowed only one request to be
+        outstanding at a time on a given TCP connection. HTTP/1.1 (<xref target="HTTP11"/>)
+        added request pipelining, but this only partially addressed request
+        concurrency and still suffers from application-layer head-of-line
+        blocking. Therefore, HTTP/1.0 and HTTP/1.1 clients that need to make
+        many requests use multiple connections to a server in order to achieve
+        concurrency.
       </t>
       <t>
         Furthermore, HTTP fields are often repetitive and verbose, causing unnecessary
@@ -95,7 +98,7 @@
         The resulting protocol is more friendly to the network because fewer TCP connections can
         be used in comparison to HTTP/1.x. This means less competition with other flows and
         longer-lived connections, which in turn lead to better utilization of available network
-        capacity.
+        capacity. Note, however, that TCP head-of-line blocking cannot be addressed by this protocol.
       </t>
       <t>
         Finally, HTTP/2 also enables more efficient processing of messages through use of binary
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2bis.xml
```diff
@@ -95,7 +98,7 @@
         The resulting protocol is more friendly to the network because fewer TCP connections can
         be used in comparison to HTTP/1.x. This means less competition with other flows and
         longer-lived connections, which in turn lead to better utilization of available network
-        capacity.
+        capacity. Note, however, that TCP head-of-line blocking cannot be addressed by this protocol.
```
martinthomson, line 101:
```suggestion
        capacity. Note, however, that TCP head-of-line blocking is not addressed by this protocol.
```

draft-ietf-httpbis-http2bis.xml
```diff
@@ -67,16 +67,19 @@
...
+        (HTTP) is linked to how each version of HTTP uses the underlying
+        transport, and the conditions under which the transport operates.
       </t>
       <t>
-        In particular, HTTP/1.0 allowed only one request to be outstanding at a time on a given TCP
-        connection. HTTP/1.1 added request pipelining, but this only partially addressed request
-        concurrency and still suffers from head-of-line blocking. Therefore, HTTP/1.0 and HTTP/1.1
-        clients that need to make many requests use multiple connections to a server in order to
-        achieve concurrency and thereby reduce latency.
+        In particular, HTTP message concurrency can reduce latency and improve
```
martinthomson, line 75:
```suggestion
        Making multiple concurrent requests can reduce latency and improve
```

draft-ietf-httpbis-http2bis.xml
```diff
@@ -67,16 +67,19 @@
...
+        application performance. HTTP/1.0 allowed only one request to be
+        outstanding at a time on a given TCP connection. HTTP/1.1 (<xref target="HTTP11"/>)
+        added request pipelining, but this only partially addressed request
+        concurrency and still suffers from application-layer head-of-line
+        blocking. Therefore, HTTP/1.0 and HTTP/1.1 clients that need to make
+        many requests use multiple connections to a server in order to achieve
+        concurrency.
```
martinthomson, line 82:
```suggestion
        blocking. Therefore, HTTP/1.0 and HTTP/1.1 clients use multiple connections
        to a server to make concurrent requests.
```

martinthomson (commented):
Thanks Lucas.  Some small suggestions only.

martinthomson (approved):

draft-ietf-httpbis-http2bis.xml
```diff
@@ -67,16 +67,18 @@
     <section anchor="intro">
       <name>Introduction</name>
       <t>
-        The Hypertext Transfer Protocol (HTTP) is a wildly successful protocol. However, the way
-        HTTP/1.1 uses the underlying transport (<xref target="HTTP11"/>) has several
-        characteristics that have a negative overall effect on application performance today.
+        The performance of application using the Hypertext Transfer Protocol
```
Lukasa, line 70:
```suggestion
        The performance of applications using the Hypertext Transfer Protocol
```

Lukasa (approved):
One change and I'm happy

LPardue:
thanks both, good suggestions

PR #907, opened 2021-08-19
Avoid previous versions, use RFC7540

### LINKED ISSUE #904: Previous version*s*
At two points in the document, it refers to "previous version**s** of this document," which I take to mean RFC 7540, since we don't discuss differences from previous drafts when we publish the RFC.  Is there a reason not to simply say "RFC 7540" in these places, similar to the discussion in 5.3?

Lukasa:
Resolves #904.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -538,7 +538,7 @@ HTTP Frame {
         </t>
         <aside>
           <t>
-            Note: Previous versions of this specification used the term "header block" in place of
+            Note: <xref target="RFC7540">RFC7540</xref> used the term "header block" in place of
             the more generic "field block".
           </t>
         </aside>
@@ -4383,7 +4383,7 @@ cookie: e=f
       <section anchor="iana-h2c">
         <name>The h2c Upgrade Token</name>
         <t>
-         Previous versions of this document (<xref target="RFC7540" section="11.8"/>) registered an upgrade
+         <xref target="RFC7540" section="11.8">RFC7540</xref> registered an upgrade
          token. This capability has been removed: see <xref target="versioning"/>.
         </t>
       </section>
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2bis.xml
```diff
@@ -538,7 +538,7 @@ HTTP Frame {
         </t>
         <aside>
           <t>
-            Note: Previous versions of this specification used the term "header block" in place of
+            Note: <xref target="RFC7540">RFC7540</xref> used the term "header block" in place of
```
martinthomson, line 541:
```suggestion
            Note: <xref target="RFC7540">RFC 7540</xref> used the term "header block" in place of
```
In prose, the name has a space (it should be non-breaking, but I don't tolerate `&#xa0;` very well.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -4383,7 +4383,7 @@ cookie: e=f
       <section anchor="iana-h2c">
         <name>The h2c Upgrade Token</name>
         <t>
-         Previous versions of this document (<xref target="RFC7540" section="11.8"/>) registered an upgrade
+         <xref target="RFC7540" section="11.8">RFC7540</xref> registered an upgrade
```
martinthomson, line 4386:
```suggestion
         <xref target="RFC7540" section="11.8"/> registered an upgrade
```
I don't know what your version would have rendered as.  If you want to try it out and share, then maybe try that.  With a space, of course.

martinthomson (approved):

MikeBishop (approved):

draft-ietf-httpbis-http2bis.xml
Lukasa: I don’t have a proper computer on me and I’m too lazy to spin up a cloud VM so I’m just gonna do it your way.

PR #909, opened 2021-08-19
Update core references

reschke:


martinthomson:
Rapidly moving target.

PR #910, opened 2021-08-19
Use absolute-path

martinthomson:
Thanks @wtarreau for find this little discrepancy.


wtarreau:
It looks so simple each time you do it for me, Martin :-)
Yes that completely solves the problem, thank you! Then I'm going to change my code to relax this now.

Lukasa (approved):

PR #914, opened 2021-08-19
extending H2: make ref to settings more generic

reschke:
Which changes

> Extensions are permitted to use new frame types (Section 4.1), new settings (Section 6.5.2), or new error codes (Section 7).

to

> Extensions are permitted to use new frame types (Section 4.1), new settings (Section 6.5), or new error codes (Section 7).


Lukasa (approved):

martinthomson:
The new reference is to the frame type (not the settings themselves), so I might have preferred the original, but I don't mind.

PR #919, opened 2021-08-20
More regarding erratum 6309

### LINKED ISSUE #915: Stream State Machine and Extension Frames
As discussed in [Erratum 6309](https://www.rfc-editor.org/errata/eid6309), the current language of 5.1 prohibits *all* frames on closed streams other than the ones indicated.  It is not clear whether the requirement in Section 5.5 constitutes "more specific guidance."  My suggestion in the erratum was to scope "any frame" to "any frame defined in this document" where we might allow an extension frame to appear, since that was the minimal change to the RFC 7540 text, but the editors might choose a broader rewording.

martinthomson:
We could try to make every statement precise.  We could add "of the
frames defined in this document" to a bunch of places.  Or we could make
a (better) blanket statement.  That is what this does.

Closes #915.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -908,11 +908,18 @@ HTTP Frame {
           </dd>
         </dl>
         <t>
-          In the absence of more specific guidance elsewhere in this document, implementations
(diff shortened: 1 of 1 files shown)


Lukasa (approved):
I think this is fine.

MikeBishop (approved):

PR #926, opened 2021-08-20
Contains is not set

### LINKED ISSUE #916: Frame "contains" an END_STREAM flag
(Warning, I may have reached peak pedantry.)

In Section 5.1, there are two references to frames which "contain" an END_STREAM flag. All HEADERS, CONTINUATION, and DATA frames contain this flag, which might be set or not.  Most other references in the document are to the flag being set, which is more precise.

There are also two references to frames "bearing" END_STREAM, which could be taken either way.

MikeBishop:
Fixes #916.


Lukasa (approved):

Lukasa:
Going to let @martinthomson have a look at this as well.

PR #928, opened 2021-08-20
Padding text on HEADERS

### LINKED ISSUE #927: Padding constraint on HEADERS
[Section 6.2](https://httpwg.org/http2-spec/draft-ietf-httpbis-http2bis.html#name-headers) says:

> The total number of padding octets is determined by the value of the Pad Length field. Padding that exceeds the size remaining for the field block fragment MUST be treated as a PROTOCOL_ERROR.

I think the intended reading is that when the parser reaches the point that only the FBF and the Padding remain, the length of Padding is given by Pad Length and the fragment is the balance.  If Pad Length exceeds the remaining length of the frame, it's an error.

There is a less-likely reading also allowed by the current phrasing, that Pad Length can't be more than half the remaining bytes (else Padding would exceed the size of the field block fragment).  That's very unlikely to be the intended reading, but we should be clear here.

I'd suggest simply taking the note from the end of DATA and putting it here verbatim.

MikeBishop:
Fixes #927 by copying the text from DATA

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -1546,10 +1546,17 @@ HEADERS Frame {
           The HEADERS frame changes the connection state as described in <xref target="FieldBlock"/>.
         </t>
         <t>
-          The total number of padding octets is determined by the value of the Pad Length field.  Padding that exceeds the size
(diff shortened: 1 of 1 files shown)


Lukasa (approved):
LGTM, will let @martinthomson take a read as well.

PR #930, opened 2021-08-20
Nits through S7

MikeBishop:
Nits through Section 7; more to follow.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -895,7 +895,7 @@ HTTP Frame {
             <t>
               An endpoint can perform this minimal processing for all streams that are in the
               "closed" state.  Endpoints MAY use other signals to detect that a peer has received
-              the frames that caused the stream to become "closed" and treat receipt of any frame other
+              the frames that caused the stream to enter the "closed" state and treat receipt of any frame other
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2bis.xml
```diff
@@ -895,7 +895,7 @@ HTTP Frame {
             <t>
               An endpoint can perform this minimal processing for all streams that are in the
               "closed" state.  Endpoints MAY use other signals to detect that a peer has received
-              the frames that caused stream to become "closed" and treat receipt of any frame other
+              the frames that caused the stream to become "closed" and treat receipt of any frame other
```
MikeBishop, line 898: I'm also ambivalent about quoting "closed" here, since we usually only quote when talking about 'caused the stream to enter the "closed" state'

Lukasa: Why not replace ‘become “closed”’ with ‘caused the stream to enter the “closed” state’ directly? It’s more verbose but it has the advantage of being unambiguous about what ‘become “closed”’ means.

Lukasa (approved):

draft-ietf-httpbis-http2bis.xml
martinthomson:
```suggestion
              the frames that caused the stream to enter the "closed" state and treat receipt of any frame other
```

martinthomson (approved):
I'll make that one change.

PR #934, opened 2021-08-23
Ciphers were listed in 2015.

### LINKED ISSUE #897: Was the cipher suite list re-assembled?
There's a note that says 

> Note: This list was assembled from the set of registered TLS cipher suites at the time of writing.

Would be be more accurate to say this was assembled for RFC 7540, or is it fresh?

martinthomson:
Closes #897.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -5057,15 +5057,11 @@ cookie: e=f
         <li>TLS_PSK_WITH_AES_128_CCM_8</li>
         <li>TLS_PSK_WITH_AES_256_CCM_8</li>
       </ul>
-      <aside>
-        <t>Note:
-          This list was assembled from the set of registered TLS cipher suites at the time of
-          writing.  This list includes those cipher suites that do not offer an ephemeral key
-          exchange and those that are based on the TLS null, stream, or block cipher type (as
-          defined in <xref target="TLS12" section="6.2.3"/>).  Additional cipher suites with
-          these properties could be defined; these would not be explicitly prohibited.
-        </t>
-      </aside>
+      <aside><t>Note: This list was assembled from the set of registered TLS cipher suites when
+      <xref target="RFC7540"/> was developed.  This list includes those cipher suites that do not
+      offer an ephemeral key exchange and those that are based on the TLS null, stream, or block
+      cipher type (as defined in <xref target="TLS12" section="6.2.3"/>).  Additional cipher suites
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2bis.xml
```diff
@@ -5057,15 +5057,11 @@ cookie: e=f
...
-        <t>Note:
-          This list was assembled from the set of registered TLS cipher suites at the time of
-          writing.  This list includes those cipher suites that do not offer an ephemeral key
-          exchange and those that are based on the TLS null, stream, or block cipher type (as
-          defined in <xref target="TLS12" section="6.2.3"/>).  Additional cipher suites with
-          these properties could be defined; these would not be explicitly prohibited.
-        </t>
-      </aside>
+      <aside><t>Note: This list was assembled from the set of registered TLS cipher suites when
+      <xref target="RFC7540"/> was developed.  This list includes those cipher suites that do not
```
Lukasa, line 5061: Presumably this was meant to be `excludes`, not `includes`, and we just missed it?

draft-ietf-httpbis-http2bis.xml
```diff
@@ -5057,15 +5057,11 @@ cookie: e=f
...
+      offer an ephemeral key exchange and those that are based on the TLS null, stream, or block
```
Lukasa, line 5062: And should this say “TLS null stream or block”?

draft-ietf-httpbis-http2bis.xml
martinthomson: Includes is correct.  The negation comes later.

draft-ietf-httpbis-http2bis.xml
martinthomson: "null" (that is, identity or `f : x -> x`) is a type of "cipher" that some suites offer. Scare quotes on "cipher" are necessary because it's exactly as bad as it sounds.

"stream", and "block" are also types of cipher.  RC4 is an example of the former, AES CBC modes the latter.

Lukasa: Wow this text is unclear. What it’s saying is that AEAD is a kind of cipher distinct from stream, block, and null? Ok I guess.

draft-ietf-httpbis-http2bis.xml
Lukasa: It comes earlier, as it turns out, but great.

Lukasa (approved):

PR #939, opened 2021-08-23
Lower-case hex

MikeBishop:
More instances of using a-f than using A-F, so I normalized the outliers to lower-case.


martinthomson (approved):
Holding this one until that text settles though.

Lukasa (approved):

PR #940, opened 2021-08-23
Remove presupposition

MikeBishop:
The existing language presupposes that the server will reset the stream upon detecting a malformed request, when in fact it only MAY do so.  The simplest fix for this text is to remove any reference to resetting the stream and restrict the advice to using an HTTP status code.


martinthomson:
Yeah, I think that I agree, but that text is actively under discussion and might change.  I'd rather not create a conflict yet.  Holding this.

Lukasa (approved):

martinthomson:
Overtaken by events.

PR #950, opened 2021-08-24
Last round of review nits

MikeBishop:


draft-ietf-httpbis-http2bis.xml
```diff
@@ -2962,7 +2962,7 @@ cookie: e=f
           </t>
           <t>
             Individual HTTP/2 requests do not carry an explicit indicator of protocol version.
-            All HTTP/2 messages implicitly have a protocol version of "2.0" (see
+            All HTTP/2 requests implicitly have a protocol version of "2.0" (see
```
MikeBishop, line 2965: Line 2979 says the same thing about responses; I was originally going to remove one or the other as a duplicate, but this section is about requests, so the smaller change is to make them complementary statements.

martinthomson (approved):

Lukasa (approved):

PR #947, opened 2021-08-24
Refer to Extended Connect

Lukasa:
Closes #941

PR #951, opened 2021-08-25
reference extended connect

### LINKED ISSUE #941: Reference to Extended CONNECT?
In the aside about HTTP/2 deliberately not supporting Upgrade, it feels worth mentioning that we later partially reversed ourselves and brought back a moral equivalent of Upgrade with a less obvious name.  Reference to RFC8441 appropriate here?

martinthomson:
I know that Cory beat me to it with #947, but I had written text.  I only forgot to make the PR.  Cory can decide which he likes better :)

Closes #941.


Lukasa (approved):
You put more things in your citation block, so it’s clearly better.

PR #959, opened 2021-08-30
stream state diagram: move legend out of artwork

reschke:


Lukasa (approved):
I’ll have to see how this renders, but this seems reasonable to me.

reschke:
FWIW, I tried to put the definition terms into "\<tt>" elements, but this causes weirdness in the HTML; apparently margin-bottom doesn't work as intended in that case. Maybe @martinthomson knows what's going on here and what needs to be fixed in xml2rfc.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -671,17 +671,25 @@ HTTP Frame {
...
+          <dt>ES</dt>
+          <dd>END_STREAM flag</dd>
+          <dt>R</dt>
+          <dd><xref target="RST_STREAM" format="none">RST_STREAM</xref> frame</dd>
+          <dt>PP</dt>
+          <dd><xref target="PUSH_PROMISE" format="none">PUSH_PROMISE</xref> frame (with implied <xref target="CONTINUATION" format="none">CONTINUATION</xref>s)</dd>
+        </dl>
+        <t>
+          Note: State transitions are for the promised stream.
+        </t>
```
martinthomson, line 692:
```suggestion
          <dd><xref target="PUSH_PROMISE" format="none">PUSH_PROMISE</xref> frame (with implied <xref target="CONTINUATION" format="none">CONTINUATION</xref> frames; state transitions are for the promised stream)</dd>
        </dl>
```

draft-ietf-httpbis-http2bis.xml
```diff
@@ -671,17 +671,25 @@ HTTP Frame {
...
+          <dt>recv</dt>
+          <dd>endpoint receives this frame</dd>
+          <dt>H</dt>
+          <dd><xref target="HEADERS" format="none">HEADERS</xref> frame (with implied <xref target="CONTINUATION" format="none">CONTINUATION</xref>s)</dd>
+          <dt>ES</dt>
+          <dd>END_STREAM flag</dd>
+          <dt>R</dt>
+          <dd><xref target="RST_STREAM" format="none">RST_STREAM</xref> frame</dd>
+          <dt>PP</dt>
+          <dd><xref target="PUSH_PROMISE" format="none">PUSH_PROMISE</xref> frame (with implied <xref target="CONTINUATION" format="none">CONTINUATION</xref>s)</dd>
```
martinthomson, line 688:
```suggestion
          <dd><xref target="PUSH_PROMISE" format="none">PUSH_PROMISE</xref> frame (with implied <xref target="CONTINUATION" format="none">CONTINUATION</xref> frames)</dd>
```

draft-ietf-httpbis-http2bis.xml
```diff
@@ -671,17 +671,25 @@ HTTP Frame {
...
-           Note: State transitions are for the promised stream.
 ]]></artwork>
         </figure>
+        <dl spacing="compact">
+          <dt>send</dt>
+          <dd>endpoint sends this frame</dd>
+          <dt>recv</dt>
+          <dd>endpoint receives this frame</dd>
+          <dt>H</dt>
+          <dd><xref target="HEADERS" format="none">HEADERS</xref> frame (with implied <xref target="CONTINUATION" format="none">CONTINUATION</xref>s)</dd>
```
martinthomson, line 682:
```suggestion
          <dd><xref target="HEADERS" format="none">HEADERS</xref> frame (with implied <xref target="CONTINUATION" format="none">CONTINUATION</xref> frames)</dd>
```

martinthomson (approved):
Seems fine.  I've expanded CONTINUATIONS -> CONTINUATION frames and moved the note up to the definition of `PP` as it relates to that only.

martinthomson:
OK, after some tweaks, I've managed to work out what is going on.  If you want a lesson in leading text and inline layout, I might be able to give you more information than I ever wanted.  I've updated the style sheet to allow for `<tt>` or `<code>` in `<dt>`, as well as fixed some other minor layout issues.

(The short version is that changing the font causes the line to become taller, which breaks the rather fragile layout we use for `<dl>`.  The problem is that the different font has different ascender/descender heights, so when the text baseline is aligned, the extra space ends up being differently balanced to the surrounding text; the CSS layout then makes the line taller to allow for the height of all included fonts.  At that point, the floats that contain `<dt>` no longer line up with the lines that contain `<dd>`.)

Issue #905, opened 2021-08-19
Remaining corner cases between Host and :authority


wtarreau:
Adding here the link to my yesterdays report to help with tracking:

https://lists.w3.org/Archives/Public/ietf-http-wg/2021JulSep/0237.html

I can also paste the contents and/or open one or more issues if needed, though I don't think it is needed for now.

daurnimator:
I think #768 is related.

icing:
In hindsight, it is easy to see that we need an "HTTP info set", e.g. the pure semantics, and from that derive protocol version serializations. As things went, implementations are now seeking ways to make their own  internal representation of semantics. As @wtarreau describes for haproxy.

The same is happening in httpd, where one carefully tries to disect the cases that apply in general from the ones specific to a version. This is tricky for backward compatible releases.

The worst, to me, seem to be rules in standards to specify "from another http version". That seems to imply an "older, existing" version. Will it apply to a future http version? Who knows? 

But this question needs an answer soon, as implementations will need to cope with QUIC traffic incoming and forwarding it to another  H2 server. I believe implementations will need to 

 * define their own internal representation of HTTP semantics
 * how to map incoming traffic to it
 * sift through the rules in the standard and determine which can be applied safely to the generic semantics and which need to stay at "if this was originally from version x.y"

Standard revisions could help here with clarifying the "from other httpd versions" to either apply in general to http semantics or specifically to a range of HTTP versions. It's probably not easy.

mnot:
Willy, we need to be careful here -- this is a pretty broad issue, touching many points and a fair amount of history. We can't reopen HTTP-core right now, so we're constrained in how well we can integrate things here.

Looking at what you bring up, I think we can make some editorial improvements here; I've suggested some in #961. There's a very fine line to walk here between respecting the abstraction that semantics provides and making it useful to implementers.

The other thing that might be worth discussing is this requirement:

> An intermediary that forwards a request to HTTP/2 MUST retain any Host header field, even if an authority is part of control data.

My recollection is that this was advocated as helping in the reconstruction of the request-line, for purposes like bot detection, debugging, etc.; it gives fidelity to what can be put on the wire in HTTP/1 (although whether that's a good goal is another discussion -- as I've complained elsewhere, h1, Host and authority are not necessarily... sane).

[...]

PR #961, opened 2021-09-01
Editorial suggestion

mnot:
For #905.


martinthomson (commented):
This looks right.

We're losing a few `<tt>` elements here, I will need to check, but it might be necessary to add those back in.  I can do that though.

Lukasa (approved):

PR #964, opened 2021-09-02
Don't justify lowercase requirement

### LINKED ISSUE #963: lowercase field names: movivation vs requirement
> To improve efficiency and interoperability, field names MUST be
converted to lowercase when constructing an HTTP/2 message. 

I think this is somewhat misleading, it just provides the motivation
why the lowercase format was introduced initially. I would just remove
the sentence and potentially insert a note about lower-casing into the
field name validity statements.

martinthomson:
Just state it. No need to justify it so (especially as this invites
questions about the claimed benefits).

This is less than Julian wanted from #963, but I think that it is
important to include a requirement like this very clearly and it isn't
really said anywhere else.  At least not directly.

Closes #963.


Lukasa (approved):
Agreed, I don't think the spec needs to defend itself on this choice.

PR #967, opened 2021-09-02
Remove old registrations

### LINKED ISSUE #960: IANA requests seem to be incorrect for a -bis
The current IANA Considerations for H2 bis has several requests to IANA that seem to be appropriate for the original document, but not a bis now that the registries are already filled out. Shouldn't these be reworded to say "this document updates..." or explain that the references for registries should point to this document, rather than the old RFC?

`This document establishes a registry for HTTP/2 frame type codes.`

martinthomson:
As noted in #960, we don't need to redo these.  A backward reference will do.

This cuts quite a bit of XML out of the document, which I'm happy about.

Closes #960.


reschke (changes requested):
I don't think this is a good idea. At the end of the day, the IANA registry should point to the updated document.

mnot:
@reschke I disagree; IANA hold the registration data, not the document. My understanding is that templates need only appear in the originally registering documents; if there's doubt about that, we should check with the IESG and IANA.

reschke:
That is true, but the registrations *point* to specs. The IANA Considerations need to take care of updating these references.

martinthomson:
That's a good point; I will shelve that and change to adding a note that says "references to RFC 7540 for X, Y, and Z are updated to point to this document".

mnot:
> That is true, but the registrations point to specs. The IANA Considerations need to take care of updating these references.

Absolutely. It just doesn't need to list them individually; a simple statements such as "update the references to RFCxxxx to this document in Registry Foo" will suffice.

reschke:
> Absolutely. It just doesn't need to list them individually; a simple statements such as "update the references to RFCxxxx to this document in Registry Foo" will suffice.

If that's all which is necessary, yes. Hint: section numbers might change, and you don't want to handle that during AUTH48.

PR #968, opened 2021-09-14
Host and :authority must agree

### LINKED ISSUE #905: Remaining corner cases between Host and :authority

martinthomson:
This makes a few changes, restricting things further than before.  For
the most part, this removes an allowance in the original specification
that had Host and :authority potentially differing.  The goal of that
was - from memory - to preserve some of the inherent quirks in HTTP/1.1.
That turns out to be more of a liability than an asset and far less
important now that we have a more formal understanding of the structure
of requests.

Closes #905.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -2951,9 +2951,9 @@ cookie: e=f
               <t>
                 An intermediary that forwards a request received in HTTP/2 via HTTP/1.1 MUST set the
                 <tt>Host</tt> field in the forwarded request to the value from the
-                <tt>:authority</tt> pseudo-header field, unless it changes the request target. This
-                replaces any existing <tt>Host</tt> field to avoid potential vulnerabilities in HTTP
-                routing.
+                <tt>:authority</tt> pseudo-header field, unless the intermediary also changes the
+                request target. This replaces any existing <tt>Host</tt> field to avoid potential
+                vulnerabilities in HTTP routing.
               </t>
               <t>
                 An intermediary that forwards a request over HTTP/2 MAY retain any <tt>Host</tt>
@@ -5184,6 +5184,9 @@ cookie: e=f
         <li>
           Connection-specific header fields - which are prohibited - are more precisely and
           comprehensively identified.
+        </li>
+        <li>
+          <tt>Host</tt> and <tt>:authority</tt> are no longer permitted to disagree.
         </li>
       </ul>
     </section>
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -2953,7 +2953,7 @@ cookie: e=f
                 source of this information; see <xref target="HTTP" section="7.2"/>.
               </t>
               <t>
-                An intermediary that needs to produce a <tt>Host</tt> header field (which might be
+                An intermediary that needs to generate a <tt>Host</tt> header field (which might be
                 necessary to construct an HTTP/1.1 request) MUST use the value from the <tt>:authority</tt> 
                 pseudo-header field as the value of the <tt>Host</tt> field,
                 unless the intermediary also changes the request target. This replaces any existing
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -2936,7 +2936,7 @@ cookie: e=f
               </t>
               <t>
                 Clients MUST NOT generate a request with a <tt>Host</tt> header field that differs
-                from the <tt>:authority</tt> pseudo-header field, when compared byte-for-byte.  A
+                from the <tt>:authority</tt> pseudo-header field.  A
                 server SHOULD treat a request as malformed if it contains a <tt>Host</tt> header
                 field that identifies a different entity to the <tt>:authority</tt> pseudo-header
                 field.  The values of fields need to be normalized to compare them (see <xref
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -2935,15 +2935,28 @@ cookie: e=f
                 information to convey (in which case it MUST NOT generate :authority).
               </t>
               <t>
+                Clients MUST NOT generate a request with a <tt>Host</tt> header field that differs
+                from the <tt>:authority</tt> pseudo-header field.  A server MAY treat a request as
+                malformed if it contains a <tt>Host</tt> header field that is different from the
+                value of the <tt>:authority</tt> pseudo-header field.
+              </t>
+              <t>
                 An intermediary that forwards a request over HTTP/2 MUST construct an
                 <tt>:authority</tt> pseudo-header field using the authority information from the
                 control data of the original request, unless the the original request's target URI
                 does not contain authority information (in which case it MUST NOT generate
-                <tt>:authority</tt>). Note that the Host header field is not the sole source of this
-                information; see <xref target="HTTP" section="7.2"/>.
+                <tt>:authority</tt>). Note that the <tt>Host</tt> header field is not the sole
+                source of this information; see <xref target="HTTP" section="7.2"/>.
               </t>
               <t>
-                An intermediary that forwards a request over HTTP/2 MUST retain any <tt>Host</tt>
+                An intermediary that forwards a request received in HTTP/2 via HTTP/1.1 MUST set the
+                <tt>Host</tt> field in the forwarded request to the value from the
+                <tt>:authority</tt> pseudo-header field, unless it changes the request target. This
+                replaces any existing <tt>Host</tt> field to avoid potential vulnerabilities in HTTP
+                routing.
+              </t>
+              <t>
+                An intermediary that forwards a request over HTTP/2 MAY retain any <tt>Host</tt>
                 header field.
               </t>
               <t>
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -2953,11 +2953,11 @@ cookie: e=f
                 source of this information; see <xref target="HTTP" section="7.2"/>.
               </t>
               <t>
-                An intermediary that forwards a request received in HTTP/2 via HTTP/1.1 MUST set the
-                <tt>Host</tt> field in the forwarded request to the value from the
-                <tt>:authority</tt> pseudo-header field, unless the intermediary also changes the
-                request target. This replaces any existing <tt>Host</tt> field to avoid potential
-                vulnerabilities in HTTP routing.
+                An intermediary that needs to produce a <tt>Host</tt> header field (which might be
+                necessary to construct an HTTP/1.1 request) MUST set the <tt>Host</tt> field in the
+                forwarded request to the value from the <tt>:authority</tt> pseudo-header field,
+                unless the intermediary also changes the request target. This replaces any existing
+                <tt>Host</tt> field to avoid potential vulnerabilities in HTTP routing.
               </t>
               <t>
                 An intermediary that forwards a request over HTTP/2 MAY retain any <tt>Host</tt>
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -2936,9 +2936,13 @@ cookie: e=f
               </t>
               <t>
                 Clients MUST NOT generate a request with a <tt>Host</tt> header field that differs
-                from the <tt>:authority</tt> pseudo-header field.  A server MAY treat a request as
-                malformed if it contains a <tt>Host</tt> header field that is different from the
-                value of the <tt>:authority</tt> pseudo-header field.
+                from the <tt>:authority</tt> pseudo-header field, when compared byte-for-byte.  A
+                server SHOULD treat a request as malformed if it contains a <tt>Host</tt> header
+                field that identifies a different entity to the <tt>:authority</tt> pseudo-header
+                field.  The values of fields need to be normalized to compare them (see <xref
+                target="RFC3986" section="6.2"/>).  An origin server can apply any normalization
+                method, whereas other servers MUST perform scheme-based normalization (see <xref
+                target="RFC3986" section="6.2.3"/>) of the two fields.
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2bis.xml
```diff
@@ -2934,16 +2934,29 @@ cookie: e=f
...
                 does not contain authority information (in which case it MUST NOT generate
-                <tt>:authority</tt>). Note that the Host header field is not the sole source of this
-                information; see <xref target="HTTP" section="7.2"/>.
+                <tt>:authority</tt>). Note that the <tt>Host</tt> header field is not the sole
+                source of this information; see <xref target="HTTP" section="7.2"/>.
+              </t>
+              <t>
+                An intermediary that forwards a request received in HTTP/2 via HTTP/1.1 MUST set the
+                <tt>Host</tt> field in the forwarded request to the value from the
+                <tt>:authority</tt> pseudo-header field, unless it changes the request target. This
```
mnot, line 2954: 'it' is ambiguous here - are you referring to the intemediary, or the act of setting the field?

draft-ietf-httpbis-http2bis.xml
```diff
@@ -2934,16 +2934,29 @@ cookie: e=f
                 pseudo-header field to convey authority information, unless there is no authority
                 information to convey (in which case it MUST NOT generate :authority).
               </t>
+              <t>
+                Clients MUST NOT generate a request with a <tt>Host</tt> header field that differs
+                from the <tt>:authority</tt> pseudo-header field.  A server MAY treat a request as
```
mnot, line 2939: I'm not sure I get your reasoning for MAY here instead of something stronger. It seems that this is a security issue, so we should have stronger reasoning for optionality here. Personally I think this should be at least SHOULD, if not MUST.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -2934,16 +2934,29 @@ cookie: e=f
...
+                malformed if it contains a <tt>Host</tt> header field that is different from the
```
mnot, line 2940: What is 'different' here? If one is present and one isn't, is that different? What if they have case differences? Etc.

mnot (commented):
Some things to think about.

Also, it seems like something should be listed in 'Changes from...'?

draft-ietf-httpbis-http2bis.xml
martinthomson: The reasoning is that old implementations won't apply this rule (so it can't be MUST) and some might choose not to (as it requires an extra comparison).  I'm OK with "SHOULD", but "MAY" seemed safer.

draft-ietf-httpbis-http2bis.xml
martinthomson: I didn't specify that deliberately, allowing someone to treat "example.com" and "example.com:443" and "Example.com" each as different to the other.

Would it be best to point that out?

> Servers are not required to normalize these fields before comparing them, which means that clients that produce both fields need to make the values byte-for-byte identical.

draft-ietf-httpbis-http2bis.xml
martinthomson: Ack, fixed.

draft-ietf-httpbis-http2bis.xml
mnot: IMO if there's significant security impact, it can be a MUST - especially since this is about recipient behaviour, not sender behaviour.

draft-ietf-httpbis-http2bis.xml
mnot: or just add 'when their values are compared, byte-for-byte'

wtarreau: They should not be compared byte-for-byte, we did this mistake recently and broke a few clients :-)
RFC3986#6.2.3 explains the scheme-based normalization, that works particularly well for H2 (in short, drop ":80" on "http" and ":443" on "https").

I think I'm fine with a SHOULD though. As Stefan (I think) mentioned, most implementations currently do not perform the comparison and a MUST would instantly make them non-compliant.

HTTP/1.1 messaging proceeds differently (in 3.2). It enumerates what a client must send, then says that a recipient MUST reject as badreq any request with invalid or missing Host. This leaves a bit of gray area about what you consider as invalid but does mandate that *some* checks are performed. I'm not sure I like this approach better, but I wanted to note it if that can fuel the discussion.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -2934,16 +2934,29 @@ cookie: e=f
...
                 <tt>:authority</tt> pseudo-header field using the authority information from the
                 control data of the original request, unless the the original request's target URI
                 does not contain authority information (in which case it MUST NOT generate
-                <tt>:authority</tt>). Note that the Host header field is not the sole source of this
-                information; see <xref target="HTTP" section="7.2"/>.
+                <tt>:authority</tt>). Note that the <tt>Host</tt> header field is not the sole
+                source of this information; see <xref target="HTTP" section="7.2"/>.
+              </t>
+              <t>
+                An intermediary that forwards a request received in HTTP/2 via HTTP/1.1 MUST set the
```
wtarreau, line 2952: I'm still embarrassed by the normative language involving H1 here. What about this variant:

     An intermediary that needs to produce a Host header field (e.g. to translate an HTTP/2 request to HTTP/1.1) MUST set the ...

draft-ietf-httpbis-http2bis.xml
Lukasa: I definitely think we should leave a breadcrumb to, say, scheme-based normalization, as well as to the relevant bit of the -semantics draft ([§ 7.2 Host and Authority](https://www.ietf.org/archive/id/draft-ietf-httpbis-semantics-19.html#name-host-and-authority)).

draft-ietf-httpbis-http2bis.xml
Lukasa: I agree with @mnot. I also think that for _senders_ we can safely say that the fields MUST be byte-for-byte identical.

draft-ietf-httpbis-http2bis.xml
martinthomson: The problem with citing RFC 3986 is that it doesn't actually specify anything in Section 6.2.  It's all examples.

Section 7.2 of -semantics is similarly unhelpful.

However, as the server is authoritative, perhaps we can rely on its own definitions of authority and say if the fields identify different authorities - by its own definition - it can reject the request.  Would that work?

Lukasa: I guess we'd apply the same guidance to intermediaries? I think that's probably acceptable from where I'm sitting.

martinthomson: OK, I've worked something out.  It's pretty gross though.  Let me know what you think.

draft-ietf-httpbis-http2bis.xml
martinthomson: Thanks, that's a good suggestion.  I've taken it.

Lukasa (approved):
I think this is well-constructed.

wtarreau:
I think that's good as well. Let's have @gregw and Stefan have a look in order to avoid further round trips.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -2934,16 +2934,33 @@ cookie: e=f
                 pseudo-header field to convey authority information, unless there is no authority
                 information to convey (in which case it MUST NOT generate :authority).
               </t>
+              <t>
+                Clients MUST NOT generate a request with a <tt>Host</tt> header field that differs
+                from the <tt>:authority</tt> pseudo-header field, when compared byte-for-byte.  A
+                server SHOULD treat a request as malformed if it contains a <tt>Host</tt> header
+                field that identifies a different entity to the <tt>:authority</tt> pseudo-header
+                field.  The values of fields need to be normalized to compare them (see <xref
```
gregw, line 2942: Does "of fields" apply to all fields, all pseudo fields or just :authority fields?

Does the normalization "need" apply to the byte-for-byte comparison above ie. do MUST they only be byte-for-byte identical after normalization?
  
Is a "need" a MUST or a SHOULD?

draft-ietf-httpbis-http2bis.xml
```diff
@@ -2934,16 +2934,33 @@ cookie: e=f
...
                 does not contain authority information (in which case it MUST NOT generate
-                <tt>:authority</tt>). Note that the Host header field is not the sole source of this
-                information; see <xref target="HTTP" section="7.2"/>.
+                <tt>:authority</tt>). Note that the <tt>Host</tt> header field is not the sole
+                source of this information; see <xref target="HTTP" section="7.2"/>.
+              </t>
+              <t>
+                An intermediary that needs to produce a <tt>Host</tt> header field (which might be
+                necessary to construct an HTTP/1.1 request) MUST set the <tt>Host</tt> field in the
+                forwarded request to the value from the <tt>:authority</tt> pseudo-header field,
```
gregw, line 2958: Rather than "MUST set the Host field...", it would be better to say just "MUST use the value from...."

gregw (approved):
The changes are good.... but I have a couple of niggles about cumbersome wording... not that I have better suggestions.

draft-ietf-httpbis-http2bis.xml
MikeBishop: In this instance, I think "need" is a statement of fact -- these are hostnames, therefore not case-sensitive, therefore they need to be normalized to perform a useful comparison.  No need for RFC2119 language here.  However, that implies that "when compared byte-for-byte" above is wrong.

draft-ietf-httpbis-http2bis.xml
MikeBishop: While I agree that the answer to "use it for what?" should be clear from context, better to be explicit.

MikeBishop (approved):

draft-ietf-httpbis-http2bis.xml
gregw: Sorry I should have been less lazy:
```suggestion
                necessary to construct an HTTP/1.1 request) MUST use the value from the <tt>:authority</tt> 
                pseudo-header field as the value of the <tt>Host</tt> field,
```

draft-ietf-httpbis-http2bis.xml
martinthomson: We're placing a stronger requirement on the sender than the receiver.  The receiver requirements are more lax because we want to allow for folks like @icing who might want to avoid extra work to do so (either because they are happy that their h2 implementation is otherwise OK and don't want to open the patient to add extra checks that will mostly just slow them down).  We don't need to see every server doing this rejection, just a few.  Clients will learn not to put different values in.

That said, I think we can say "identical" in the first part without resorting to "byte-for-byte".  "byte-for-byte" is just the most obvious and easiest way to ensure that the values are the same.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -2934,16 +2934,33 @@ cookie: e=f
                 pseudo-header field to convey authority information, unless there is no authority
                 information to convey (in which case it MUST NOT generate :authority).
               </t>
+              <t>
+                Clients MUST NOT generate a request with a <tt>Host</tt> header field that differs
+                from the <tt>:authority</tt> pseudo-header field, when compared byte-for-byte.  A
```
martinthomson, line 2939:
```suggestion
                from the <tt>:authority</tt> pseudo-header field.  A
```

wtarreau:
Yeah much better now. A coworker also told me he was bothered by the "byte-for-byte" which is now gone. I've reread all of it from https://github.com/martinthomson/http2v2/blob/af9a40279c465e0b6d7994710affe94ef5228109/draft-ietf-httpbis-http2bis.xml#L2900 to see it in context and I find that it flows nicely, details the protocol particularities and traps to care about without stepping over other protocol versions. So that's a definite +1 from me!

I'm finding something confusing two paragraphs later however: it's said `Note that request targets for CONNECT or asterisk-form OPTIONS requests never include authority information` which contradicts the CONNECT section saying `A CONNECT header section is constructed ... The :authority pseudo-header field contains the host and port to connect to`.

And I'm sure I participated to that but don't remember how we ended up on this one :-/

draft-ietf-httpbis-http2bis.xml
icing: > ...we want to allow for folks like @icing who might want to avoid extra work to do so...

That's very kind and heart-warming.😌

MikeBishop:
@wtarreau, that's because the `:authority` pseudo-header in a CONNECT request is the target of the connection, not the proxy you're asking to handle the request.  It's not that the field is absent, but that it means something different in that context.

wtarreau:
@MikeBishop OK but regardless it still looks like recent clients are trying hard to send the exact same Host as the :authority in CONNECT (typically including :443).  It's unclear to me why the target of a proxied CONNECT is not an authority while the target of a proxied GET is an authority then.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -2934,16 +2934,33 @@ cookie: e=f
...
                 <tt>:authority</tt> pseudo-header field using the authority information from the
                 control data of the original request, unless the the original request's target URI
                 does not contain authority information (in which case it MUST NOT generate
-                <tt>:authority</tt>). Note that the Host header field is not the sole source of this
-                information; see <xref target="HTTP" section="7.2"/>.
+                <tt>:authority</tt>). Note that the <tt>Host</tt> header field is not the sole
+                source of this information; see <xref target="HTTP" section="7.2"/>.
+              </t>
+              <t>
+                An intermediary that needs to produce a <tt>Host</tt> header field (which might be
```
mnot, line 2956:
```suggestion
                An intermediary that needs to generate a <tt>Host</tt> header field (which might be
```

PR #970, opened 2021-09-26
Only update registries that refer to RFC 7540

martinthomson:
Not all of the references.  That would be bad.


martinthomson:
Fixed by #971.

PR #971, opened 2021-09-26
Registry names

martinthomson:
Includes changes from #970, as that hits the same paragraph.


mnot:
why is c4d3528 included?

martinthomson:
Hmm, poor rebase on my part I think.  I made #970, you made the changes there, then I made a branch for this and then rebased on top of #970.  It should go away when merged.

PR #972, opened 2021-09-27
Manually add quotes to important strings

martinthomson:
A recent xml2rfc update changed the way that the `<tt>` element was
rendered in text.  Take advantage of that by adding quotes to strings
that really do benefit from additional quoting, such as the "PRI ..."
preface string.

Most of the changes shown in a recent diff are down to losing quotes
on header field names and other protocol elements.  These do not need
to be quoted for their meaning to be clear and unambiguous.  The use
of `<tt>` for these items improves the HTML rendering, but the loss of
line noise in text renderings is a net win (at least in my opinion).


Lukasa (approved):
I agree with your reasoning, this seems like a good change.

PR #975, opened 2021-11-11
First pass cleaning up some of the typos from AD review

Lukasa:
Partially addresses some of #974. This is the low hanging fruit.

PR #976, opened 2021-11-11
Clarify some settings behaviours based on AD review

Lukasa:


draft-ietf-httpbis-http2bis.xml
```diff
@@ -1994,7 +1997,7 @@ Setting {
             other frame processing between values.  Unsupported settings MUST be ignored.  Once
             all values have been processed, the recipient MUST immediately emit a SETTINGS frame
             with the ACK flag set. Upon receiving a SETTINGS frame with the ACK flag set, the sender
-            of the altered settings can rely on the value having been applied.
+            of the altered settings can rely on the values having been applied or ignored.
```
martinthomson, line 2000: Text overlap with #975?

martinthomson (approved):

draft-ietf-httpbis-http2bis.xml
Lukasa: Ok, that makes sense. I'll update this patch to remove this text.

Issue #974, opened 2021-11-10
AD review of draft-ietf-httpbis-http2bis-05


fpalombini:
- [x] 1. -----

FP: Is the working group aware that RFC 793bis is in IESG evaluation (https://datatracker.ietf.org/doc/draft-ietf-tcpm-rfc793bis/) ? Was the choice of having a normative reference to 793 conscious, in order to avoid any delay that might come from publication of draft-ietf-tcpm-rfc793bis? (Just checking this was considered)

- [x] 2. -----

   for HTTP/2 over TLS.  The general TLS usage guidance in [TLSBCP]
   SHOULD be followed, with some additional restrictions that are
   specific to HTTP/2.

FP: Given this requirement, I would have expected to see [TLSBCP] normatively referenced, rather than informatively.

- [x] 3. -----

   layer.  The frame and stream layers are tailored to the needs of the
   HTTP protocol and server push.

FP: I would think the server push is part of the HTTP protocol, which makes this formulation "HTTP protocol and server push" confusing.

- [x] 4. -----

   Type:  The 8-bit type of the frame.  The frame type determines the

   Flags:  An 8-bit field reserved for boolean flags specific to the

FP: I would find a reference to the IANA registry useful.

- [x] 5. -----

   implementation of flow control can be difficult.  When using flow
   control, the receiver MUST read from the TCP receive buffer in a
   timely fashion.  Failure to do so could lead to a deadlock when

FP: "When using flow control" might be rephrased to indicate "when flow control limits are lower than the maximum" (or something of the sort), since if I understand correctly the capability is always used, it is just the window size that changes (effectively implementing control of flow). Also "timely fashion" - I know that it is probably hard, but it would be nice to have a more precise qualification, or at least a hint of what is timely.

- [x] 6. -----

   stream that it successfully received from its peer.  The GOAWAY frame
   includes an error code that indicates why the connection is

FP: "error code" - here as well I would have liked a reference to the IANA registry.

- [x] 7. -----

   Exclusive:  A single-bit flag.  This field is only present if the
      PRIORITY flag is set.

   Stream Dependency:  A 31-bit stream identifier.  This field is only
      present if the PRIORITY flag is set.

   Weight:  An unsigned 8-bit integer.  This field is only present if
      the PRIORITY flag is set.

FP: I would have expected to see some definition of how the fields are used. If this is defined somewhere else, a reference would be good.

- [x] 8. -----

   SETTINGS Frame {
     Length (24),
[...]

PR #981, opened 2021-11-12
Fix the flow control text

martinthomson:
Editorial change only.  The idea is to read as soon as data arrives in
the TCP buffer.

In practice, this is not necessary, but leaving things in the TCP buffer
indefinitely can deadlock.  (Some implementations of over TCP-based
protocols do this and in this case it is BAD.)

For #974.


Lukasa (approved):
Yeah, good rewording here.

Issue #990, opened 2021-11-29
tsvart review from Jörg Ott


martinthomson:
- [x] Section 5.1.1., 3rd para:

    When a stream transitions out of
the "idle" state, all streams that might have been initiated by that
peer with a lower-valued stream identifier are implicitly
transitioned to "closed". 

    Should one say "that might have been but were not initiated by that peers"
to make clear that this applies only to un-used streams?

- [x] Section 5.3.1
Maybe update the section heading to read
"Background on Priority in HTTP/2 as per RFC7540"?

- [x] Section 6.5:
The document implies that SETTINGS frames are ACKed in the order they are sent.
Should one make this explicit?

- [x] Last para of 6.5, top of page 37:
"If the sender of a SETTINGS frame does not receive an acknowledgement
within a reasonable amount of time, it MAY issue a connection error
(Section 5.4.1) of type SETTINGS_TIMEOUT."

    Should one provide some intuition on what "reasonable" could mean? An example?
It seems, an HTTP/2 sender of a SETTINGS frame could know when this was passed
to the TCP layer and may have some insights on application layer RTTs.

- [x] Section 6.9.1: Concerning flow control, should one say something about avoiding
silly windows?

- [x] Section 8.1.1, p52 1st para: Probably just me, but I am mildly confused by the
statement about messages having a non-zero content-length field but still no
data in the data frames. The reference section 6.4 of the HTTP semantics draft
didn't resolve that. 

- [x] Section 8.2.1, p53: parse error?
"When a request message violates one of these requirements, an
implementation SHOULD generate a Section 15.5.1 of 400 (Bad Request)
status code [HTTP], [...]"

- [x] Section 8.4.2, 4th para:
"A client can use the SETTINGS_MAX_CONCURRENT_STREAMS setting to limit
the number of responses that can be concurrently pushed by a server.
Advertising a SETTINGS_MAX_CONCURRENT_STREAMS value of zero disables
server push by preventing the server from creating the necessary
streams.  This does not prohibit a server from sending PUSH_PROMISE
frames; clients need to reset any promised streams that are not
wanted. 

    Confused: it disables server push, but the server can send PUSH_PROMISE frames
and then the client is supposed to reset a stream it wasn't willing to accept or
capable of accepting? Which stream number would the server legitimately put
into the PUSH_PROMISE?

- [ ] There are a few cases where I am wondering if normative language should be used.
Examples include
[...]

PR #991, opened 2021-11-29
"idle"->"closed" transition

martinthomson:
I improved the text along the lines Jörg suggested.  However, I realized
that the transition should really be in the diagram.  That's basically
impossible given the nature of the diagram, but it's worth including in
the description at least.

For #990.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -841,7 +841,12 @@ HTTP Frame {
                     Note that the <xref target="PUSH_PROMISE" format="none">PUSH_PROMISE</xref> frame is not sent on the idle
                     stream but references the newly reserved stream in the Promised Stream ID
                     field.
-                  </li>
+              </li>
+              <li>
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2bis.xml
```diff
@@ -4206,8 +4211,8 @@ cookie: e=f
             Institute</organization>
             </author>
             <date year="1981" month="September"/>
-	  </front>
-	</reference>
+    </front>
+  </reference>
```
Lukasa, line 4215: Is this whitespace change intentional?

Lukasa (approved):
This is a good clarification. One note about a whitespace change but the textual changes are :+1:

draft-ietf-httpbis-http2bis.xml
martinthomson: No, my editor does a cleanup when I close it, and I caught this.  I made the change on the main branch, so this is redundant (and wrong).

PR #997, opened 2021-11-29
PUSH_PROMISE and SETTINGS_MAX_CONCURRENT_STREAMS

martinthomson:
This is confusing; hopefully this rewrite is a little less so.

For #990.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -3297,10 +3297,12 @@ cookie: e=f
             format="none">SETTINGS_MAX_CONCURRENT_STREAMS</xref> setting to limit the number of
             responses that can be concurrently pushed by a server. Advertising a <xref
             target="SETTINGS_MAX_CONCURRENT_STREAMS"
-            format="none">SETTINGS_MAX_CONCURRENT_STREAMS</xref> value of zero disables server push
-            by preventing the server from creating the necessary streams. This does not prohibit a
-            server from sending <xref target="PUSH_PROMISE" format="none">PUSH_PROMISE</xref>
-            frames; clients need to reset any promised streams that are not wanted.
+            format="none">SETTINGS_MAX_CONCURRENT_STREAMS</xref> value of zero prevents the server
+            from open the streams necessary to push responses. However, this does not prevent the
+            server from reserving streams using <xref target="PUSH_PROMISE"
+            format="none">PUSH_PROMISE</xref> frames, because "reserved" streams do not count toward
+            the concurrent stream limit. Clients that do not with to receive pushed resources need
+            to reset any unwanted reserved streams.
           </t>
           <t>
             Clients receiving a pushed response MUST validate that either the server is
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2bis.xml
```diff
@@ -3297,10 +3297,12 @@ cookie: e=f
             format="none">SETTINGS_MAX_CONCURRENT_STREAMS</xref> setting to limit the number of
             responses that can be concurrently pushed by a server. Advertising a <xref
             target="SETTINGS_MAX_CONCURRENT_STREAMS"
-            format="none">SETTINGS_MAX_CONCURRENT_STREAMS</xref> value of zero disables server push
-            by preventing the server from creating the necessary streams. This does not prohibit a
-            server from sending <xref target="PUSH_PROMISE" format="none">PUSH_PROMISE</xref>
-            frames; clients need to reset any promised streams that are not wanted.
+            format="none">SETTINGS_MAX_CONCURRENT_STREAMS</xref> value of zero prevents the server
+            from open the streams necessary to push responses. However, this does not prevent the
```
LPardue, line 3301:
```suggestion
            from opening the streams necessary to push responses. However, this does not prevent the
```

draft-ietf-httpbis-http2bis.xml
```diff
@@ -3297,10 +3297,12 @@ cookie: e=f
...
+            server from reserving streams using <xref target="PUSH_PROMISE"
+            format="none">PUSH_PROMISE</xref> frames, because "reserved" streams do not count toward
+            the concurrent stream limit. Clients that do not with to receive pushed resources need
```
LPardue, line 3304:
```suggestion
            the concurrent stream limit. Clients that do not wish to receive pushed resources need
```

draft-ietf-httpbis-http2bis.xml
```diff
@@ -3297,10 +3297,12 @@ cookie: e=f
...
-            format="none">SETTINGS_MAX_CONCURRENT_STREAMS</xref> value of zero disables server push
-            by preventing the server from creating the necessary streams. This does not prohibit a
-            server from sending <xref target="PUSH_PROMISE" format="none">PUSH_PROMISE</xref>
-            frames; clients need to reset any promised streams that are not wanted.
+            format="none">SETTINGS_MAX_CONCURRENT_STREAMS</xref> value of zero prevents the server
+            from opening the streams necessary to push responses. However, this does not prevent the
+            server from reserving streams using <xref target="PUSH_PROMISE"
+            format="none">PUSH_PROMISE</xref> frames, because "reserved" streams do not count toward
+            the concurrent stream limit. Clients that do not wish to receive pushed resources need
+            to reset any unwanted reserved streams.
```
Lukasa, line 3305: Should we explicitly point towards `SETTINGS_ENABLE_PUSH` here? It's only a few paragraphs away in the text but this is the second time we've had a review that missed that setting, and it fits well with the text here:

> Clients that do not wish to receive pushed resources need to reset any unwanted reserved streams, or set `SETTINGS_ENABLE_PUSH` to `0` [xref whatever].

Lukasa (approved):

PR #1001, opened 2021-12-07
Sean's SECDIR review

martinthomson:
I'll need to respond via email, but these changes all seem fine.  I'm less sure about the first one than the others though, so look at that carefully.  I'm equally happy pushing back on that; Sean is right that this is probably the most important factoid in the text, but the reordering is a little awkward.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -1165,8 +1165,9 @@ HTTP Frame {
                 window.  This ensures that important control frames are not blocked by flow control.
               </li>
             <li>
-                Flow control cannot be disabled.
-              </li>
+              An endpoint can choose to disable its own flow control, but an endpoint cannot ignore
+              flow control signals from its peer.
+            </li>
             <li>
                 HTTP/2 defines only the format and semantics of the <xref target="WINDOW_UPDATE" format="none">WINDOW_UPDATE</xref>
                 frame (<xref target="WINDOW_UPDATE"/>).  This document does not stipulate how a
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -3054,8 +3054,8 @@ cookie: e=f
               <t>
                 The <tt>:authority</tt> pseudo-header field conveys the authority portion (<xref
                 target="RFC3986" section="3.2"/>) of the target URI (<xref target="HTTP"
-                section="7.1"/>). The recipient of a HTTP/2 request MUST ignore the <tt>Host</tt>
-                header field if <tt>:authority</tt> is present.
+                section="7.1"/>). The recipient of a HTTP/2 request MUST NOT use the <tt>Host</tt>
+                header field to determine the target URI if <tt>:authority</tt> is present.
               </t>
               <t>
                 Clients that generate HTTP/2 requests directly MUST use the <tt>:authority</tt>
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2bis.xml
```diff
@@ -262,21 +262,26 @@
...
       <t>
-        HTTP/2 uses the "http" and "https" URI schemes defined in <xref target="HTTP" section="4.2"/>, with the same default port numbers.  As a result,
-        implementations processing requests for target resource URIs like <tt>http://example.org/foo</tt> or <tt>https://example.com/bar</tt> are required to first discover whether the
-        upstream server (the immediate peer to which the client wishes to establish a connection)
-        supports HTTP/2.
+        HTTP/2 uses the "http" and "https" URI schemes defined in <xref target="HTTP"
+        section="4.2"/>, with the same default port numbers as HTTP/1.1 (<xref
+        target="HTTP11"/>). These URIs do not include any indication about what versions an upstream
+        server (the immediate peer to which the client wishes to establish a connection) supports
+        HTTP/2.
```
Lukasa, line 273: The last sentence here isn't right: it reads: "These URIs do not include any indication about what versions an upstream server supports HTTP/2".

Consider:

- "These URIs do not include any indication about what HTTP versions an upstream server supports."

draft-ietf-httpbis-http2bis.xml
```diff
@@ -262,21 +262,26 @@
...
       <t>
         The means by which support for HTTP/2 is determined is different for "http" and "https"
         URIs.  Discovery for "https" URIs is described in <xref target="discover-https"/>. HTTP/2
         support for "http" URIs can only be discovered by out-of-band means, and requires prior knowledge
         of the support as described in <xref target="known-http"/>.
       </t>
+      <t>
+        An HTTP/2 connection is an application-layer protocol running on top of a TCP connection
+        (<xref target="TCP"/>). The client is the TCP connection initiator.
+      </t>
```
Lukasa, line 284: I think moving this down is fine but this paragraph now reads a bit strangely. It feels like a dangling paragraph fragment, like we wanted to put this information in somewhere but couldn't find anywhere good so we jammed it here. Which I guess is accurate, but I'd like to be able to hide that.

We have defined the term "client" in the glossary above, so we don't need to redefine it here. If we want we can move the "TCP connection initiator" piece up to the glossary too. We could then move the "application-layer protocol running on top of a TCP connection" up to the overview. WDYT?

draft-ietf-httpbis-http2bis.xml
```diff
@@ -467,9 +471,9 @@ HTTP Frame {
                 An 8-bit field reserved for boolean flags specific to the frame type.
             </t>
             <t>
-                Flags are assigned semantics specific to the indicated frame type.  Unused flags are those that have
-                no defined semantics for a particular frame type, and MUST be ignored and MUST be left
-                unset (0x0) when sending.
+              Flags are assigned semantics specific to the indicated frame type.  Unused flags are
+              those that have no defined semantics for a particular frame type. Unused flags MUST be
+              ignored on receipt and MUST be left unset (0x0) when sending.
```
Lukasa, line 476: Was the decent here intentional?

martinthomson: The indent was poor before, but I "fixed" it when reflowing the text.

martinthomson:
Thanks for the review; I've tried your suggestion.  It seems to work.

Lukasa (approved):

PR #1002, opened 2021-12-15
Add 0x to Frame layout diagrams for consistency

LPardue:
This stems from the IESG review of the priorities draft, where Éric Vyncke noted the inconsistenc between prose and diagram. This change to use 0x throughout makes it consistent with HTTP/3, which helps documents that need to define frames for both versions.


Lukasa (approved):
This LGTM, but I'll let @martinthomson confirm that we're still happy to take editorial changes at this time.

LPardue:
You can theoretically take them up until the RFC is published, with vary layers of pain. Since this was caught in our IESG review, it's likely to come up in yours. 

Your editor discretion on merging this change of course.

Issue #1008, opened 2022-01-05
editorial consistency of "in octets" and "in units of octets"


kaduk:
We have a handful of places in http2bis that say "in units of octets" (3) and another handful (5) that just use "in octets".  Do we want to pick one and be consistent?

PR #1009, opened 2022-01-05
Editorial suggestions from Benjamin Kaduk's IESG Evaluation

kaduk:
of http2bis.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -1583,7 +1583,7 @@ HEADERS Frame {
   [Stream Dependency (31)],
   [Weight (8)],
   Field Block Fragment (..),
-  Padding (..255),
+  Padding (..2040),
 }
 ]]></artwork>
         </figure>
@@ -2065,7 +2065,7 @@ PUSH_PROMISE Frame {
   Reserved (1),
   Promised Stream ID (31),
   Field Block Fragment (..),
-  Padding (..255),
+  Padding (..2040),
 }
 ]]></artwork>
         </figure>
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -1483,7 +1483,7 @@ DATA Frame {
 
   [Pad Length (8)],
   Data (..),
-  Padding (..),
+  Padding (..255),
 }
 ]]></artwork>
         </figure>
@@ -1583,7 +1583,7 @@ HEADERS Frame {
   [Stream Dependency (31)],
   [Weight (8)],
   Field Block Fragment (..),
-  Padding (..),
+  Padding (..255),
 }
 ]]></artwork>
         </figure>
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2bis.xml
```diff
@@ -445,7 +445,7 @@ HTTP Frame {
           <dt>Length:</dt>
           <dd>
             <t>
-                The length of the frame payload expressed as an unsigned 24-bit integer.  Values
+                The length of the frame payload expressed as an unsigned 24-bit integer in units of octets.  Values
```
kaduk, line 448: Might need to change per #1008

draft-ietf-httpbis-http2bis.xml
```diff
@@ -1483,7 +1483,7 @@ DATA Frame {
 
   [Pad Length (8)],
   Data (..),
-  Padding (..),
+  Padding (..255),
```
kaduk, line 1486: With only one byte for pad length, the length is in effect capped here.
This is a separate commit for ease of reverting, if needed.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -1703,7 +1703,7 @@ HEADERS Frame {
           <name>PRIORITY Frame Format</name>
           <artwork type="inline"><![CDATA[
 PRIORITY Frame {
-  Length (24),
+  Length (24) = 0x5,
```
kaduk, line 1706: There were already a few places where we listed the hardcoded stream identifier in the presentation language, so I figured it was fair game for including statically known values in the presentation language, even when it's part of the generic frame header.  This is also a separate commit for ease of reverting, if needed.

draft-ietf-httpbis-http2bis.xml
```diff
@@ -2526,7 +2526,7 @@ WINDOW_UPDATE Frame {
             Prior to receiving a <xref target="SETTINGS" format="none">SETTINGS</xref> frame that sets a value for
             <xref target="SETTINGS_INITIAL_WINDOW_SIZE" format="none">SETTINGS_INITIAL_WINDOW_SIZE</xref>, an endpoint can only use the default
             initial window size when sending flow-controlled frames.  Similarly, the connection flow-control
-            window is set to the default initial window size until a WINDOW_UPDATE frame is
+            window is set based on the default initial window size until a WINDOW_UPDATE frame is
```
kaduk, line 2529: "is set to" could be misread as saying that it remains set to that fixed value even as data is sent/received, which would disable flow control until the first WINDOW_UPDATE.  That's clearly a wrong reading, but maybe this wording tweak helps.

draft-ietf-httpbis-http2bis.xml
martinthomson: No, this is a good catch, thanks.

draft-ietf-httpbis-http2bis.xml
martinthomson: I'll handle that, thanks.

martinthomson (approved):

draft-ietf-httpbis-http2bis.xml
martinthomson: Mostly...

```suggestion
  Padding (..2040),
```

kaduk: Oops, good catch.  I pushed a fixup for the other two instances.

PR #1012, opened 2022-01-07
Kaduk IESG feedback

martinthomson:
Lots of small changes here, though nothing of any real consequence, I believe.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -4926,7 +4926,9 @@ cookie: e=f
       </ul>
       <t>
         Editorial changes are also included. In particular, changes to terminology and document
-        structure are in response to updates to <xref target="HTTP">core HTTP semantics</xref>.
+        structure are in response to updates to <xref target="HTTP">core HTTP
+        semantics</xref>. Those documents now include some concepts that were first defined in RFC
+        7540, such as the 421 status code or connection coalescing.
       </t>
     </section>
     <section numbered="false">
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2bis.xml
```diff
@@ -309,9 +309,10 @@
                 document do not hold unless TLS is used; see <xref target="security"/>.
             </t>
             <t>
-                The "h2c" string was previously used as a token for use in the HTTP Upgrade mechanism's
-                Upgrade header field (<xref target="HTTP" section="7.8"/>). This usage was never widely
-                deployed, and is deprecated by this document.
+                The "h2c" string was previously used as a token for use in the HTTP Upgrade
```
mnot, line 312:
```suggestion
                The "h2c" string was previously used in the HTTP Upgrade
```

draft-ietf-httpbis-http2bis.xml
```diff
@@ -538,9 +539,10 @@ HTTP Frame {
           processes as header compression and decompression.
         </t>
         <t>
-          Field blocks carry the compressed bytes of a field section, with header sections
-          also carrying control data associated with the message in the form of <xref target="PseudoHeaderFields">pseudo-header fields</xref> that use the same format as a
-          field line.
+          A field section consists of exactly one field block.  Each field block carries the
+          compressed bytes of a single field section.  Header sections also include control data
```
Lukasa, line 543: Hmm, this section reads strangely to me. Field sections are made up of field blocks which carry the compressed bytes of field sections? This feels a little recursive.

Do we have a clearer way to express this idea?

martinthomson: I gave it a try.  I had to submit a version, so I went with this:

>          Each field block carries all of the compressed field lines of a single field section.

Lukasa: I think that would work much better.

martinthomson:
Merged manually.

PR #1017, opened 2022-01-13
Add some warnings about concatenation of URI parts

### LINKED ISSUE #1015: pseudo-header stricter validation
Before we definitely forget about it, could we please have a consideration for that one:

   https://github.com/httpwg/http2-spec/pull/936#issuecomment-910260086

There is nothing in H2 that prevents pseudo headers from being abused, and it's suggested that they have to be reassembled to make the whole request line, that is usually parsed later to extract the relevant parts, leading to the trouble reported in the portswigger article. And semantics only apply to the recomposed parts, which is too late.

I would really like that we suggest controls there so that future implementations do not get caught by this uncovered area, and the comment above proposes the minimalist controls that protect against such attacks.

martinthomson:
It's going to happen anyway, but we can at least warn people.

This doesn't specify what validation is necessary in any detail (that is
somewhat involved), it only notes that simple concatenation is almost
certainly not secure.

Closes #1015.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -3883,6 +3883,13 @@ cookie: e=f
           could be exploited by an attacker.
         </t>
         <t>
+          <xref target="HttpHeaders"/> does not include specific rules for validation of
+          pseudo-header fields.  If the values of these fields are used, additional validation is
+          necessary.  This is particularly important where <tt>:scheme</tt>, <tt>:authority</tt>,
+          and <tt>:path</tt> are combined to form a single URI string (<xref target="RFC3986"/>).
+          Simple concatenation is not secure unless the input values are fully validated.
+        </t>
+        <t>
           An intermediary can reject fields that contain invalid field names or values for other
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2bis.xml
```diff
@@ -3882,6 +3882,13 @@ cookie: e=f
           delimiters could be used to cause recipients to incorrectly interpret a message, which
           could be exploited by an attacker.
         </t>
+        <t>
+          <xref target="HttpHeaders"/> does not include specific rules for validation of
+          pseudo-header fields.  If the values of these fields are used, additional validation is
+          necessary.  This is particularly important where <tt>:scheme</tt>, <tt>:authority</tt>,
+          and <tt>:path</tt> are combined to form a single URI string (<xref target="RFC3986"/>).
```
wtarreau, line 3889: maybe add "and/or when combined with <tt>:method</tt>to rebuild a request line" since all implementations were caught with a space in that one.

martinthomson: Thanks Willy, I've added a sentence.

wtarreau:
OK that's fine by me. At least they're warned, and those who want to know more will easily find info on the subject, including these conversations. Thank you Martin!

Lukasa (approved):
LGTM.

PR #1019, opened 2022-01-24
A few small edits

martinthomson:


draft-ietf-httpbis-http2bis.xml
```diff
@@ -1565,6 +1565,13 @@ DATA Frame {
               the identified stream.  Setting this flag causes the stream to enter one of <xref target="StreamStates">the "half-closed" states or the "closed" state</xref>.
             </dd>
         </dl>
+        <aside>
+          <t>
+            Note: An endpoint that learns of stream closure after sending all data can close a
+            stream by sending a STREAM frame with a zero-length Data field and the END_STREAM flag
+            set.
```
martinthomson, line 1572: I just realized that this isn't 100% correct.  If the endpoint sends trailers, END_STREAM needs to be attached to the HEADERS frame.

Lukasa (approved):
LGTM.

PR #1020, opened 2022-02-03
Address IANA note

Lukasa:
Standard is not a status a header field can have, it turns out.

There isn't necessarily time in the process to make this change, but as a just-in-case measure I opened this. IANA has correctly updated the assignment anyway.


martinthomson (approved):

Lukasa:
No longer necessary, favoured #1021.

PR #1022, opened 2022-02-16
Pad hexadecimal literals to even digits

martinthomson:
We decided to do this for QUIC and it does improve readability and
consistency.

I did this manually, rather than trying to be clever with regexs.  I'd
appreciate a second pair of eyes.

I searched for `0x[0-9a-f](?:[0-9a-f]{2})*[^0-9a-f]` and found nothing;
same for `0x[0-9a-f]*[A-F][0-9a-f]*` (in case we had stray uppercase
strings).

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -466,21 +466,21 @@ HTTP Frame {
             <t>
               Flags are assigned semantics specific to the indicated frame type.  Unused flags are
               those that have no defined semantics for a particular frame type. Unused flags MUST be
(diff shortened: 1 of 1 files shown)


MikeBishop:
I remember where I tripped up with the regex when we did it for QUIC, so I could reconstruct it pretty quickly. I think it was something along the lines of `0x[0-9a-fA-F]([0-9a-fA-F]{2})*[^0-9a-fA-F]`.

Lukasa (approved):
LGTM

MikeBishop (approved):
Double-checked with a regex, and I don't find any missing either.

PR #1076, opened 2022-03-22
bandwidth * delay product

### LINKED ISSUE #1032: Sections 5.2.2 and 5.2.3 (comment 6)
[Sections 5.2.2](https://www.rfc-editor.org/authors/rfc9113.html#section-5.2.2) and 5.2.3:  Should "bandwidth-delay
product" be written as "bandwidth * delay product" per RFC 7323?

Original:

     Note, however, that this can lead to suboptimal use of available
     network resources if flow control is enabled without knowledge of the
     bandwidth-delay product (see [RFC7323]).

 Even with full awareness of the current bandwidth-delay product,
 implementation of flow control can be difficult.
...
 If an endpoint cannot ensure that its peer always has available flow
 control window space that is greater than the peer's bandwidth-delay
 product on this connection, its receive throughput will be limited by
 HTTP/2 flow control.

martinthomson:
This looks terrible, but it's the form that the RFC we cite uses (also,
RFC 1323).

Closes #1032.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -1223,13 +1223,13 @@ HTTP Frame {
             Deployments with constrained resources (for example, memory) can employ flow control to
             limit the amount of memory a peer can consume.  Note, however, that this can lead to
             suboptimal use of available network resources if flow control is enabled without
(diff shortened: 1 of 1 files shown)


Lukasa (approved):
Thanks, I hate it.

PR #1101, opened 2022-03-22
Apply textual changes from RFC Editors

Lukasa:
This PR contains textual changes made by the RFC editors, along with a
couple of XML changes where they were uncontroversial or necessary for
the text. I left out any change the RFC editors made that conflicted
substatially with changes we made in response to their comments.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -2481,8 +2481,8 @@ WINDOW_UPDATE Frame {
         </t>
         <t>
           WINDOW_UPDATE can be sent by a peer that has sent a frame with the END_STREAM flag set.
-          This means that a receiver could receive a WINDOW_UPDATE frame on a "half-closed (remote)"
-          or "closed" stream.  A receiver MUST NOT treat this as an error (see <xref target="StreamStates"/>).
+          This means that a receiver could receive a WINDOW_UPDATE frame on a stream in a "half-closed (remote)"
+          or "closed" state.  A receiver MUST NOT treat this as an error (see <xref target="StreamStates"/>).
         </t>
         <t>
           A receiver that receives a flow-controlled frame MUST always account for its contribution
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2bis.xml
```diff
@@ -4951,7 +4951,7 @@ cookie: e=f
       </t>
       <ul spacing="normal">
         <li>
-          Use of TLS 1.3 was defined based on RFC 8740, which this document obsoletes.
+          Use of TLS 1.3 was defined based on <xref target="RFC8740" format="default"/>, which this document obsoletes.
```
martinthomson, line 4954:
```suggestion
          Use of TLS 1.3 was defined based on <xref target="RFC8740"/>, which this document obsoletes.
```

draft-ietf-httpbis-http2bis.xml
```diff
@@ -2485,8 +2481,8 @@ WINDOW_UPDATE Frame {
         </t>
         <t>
           WINDOW_UPDATE can be sent by a peer that has sent a frame with the END_STREAM flag set.
-          This means that a receiver could receive a WINDOW_UPDATE frame on a stream in the "half-closed (remote)"
-          or "closed" state.  A receiver MUST NOT treat this as an error (see <xref target="StreamStates"/>).
+          This means that a receiver could receive a WINDOW_UPDATE frame on a "half-closed (remote)"
+          or "closed" stream.  A receiver MUST NOT treat this as an error (see <xref target="StreamStates"/>).
```
martinthomson, line 2485: I think that the original is more consistent with other text.  Though s/the/a/.

```suggestion
          This means that a receiver could receive a WINDOW_UPDATE frame on a stream in a "half-closed (remote)"
          or "closed" state.  A receiver MUST NOT treat this as an error (see <xref target="StreamStates"/>).
```

draft-ietf-httpbis-http2bis.xml
```diff
@@ -2667,7 +2663,7 @@ CONTINUATION Frame {
         </t>
         <t>
           CONTINUATION frames MUST be associated with a stream. If a CONTINUATION frame is received
-          with a Stream Identifier field of 0x00, the recipient MUST respond with a <xref target="ConnectionErrorHandler">connection error</xref> of type PROTOCOL_ERROR.
+          whose Stream Identifier field is 0x00, the recipient MUST respond with a <xref target="ConnectionErrorHandler">connection error</xref> of type PROTOCOL_ERROR.
```
martinthomson, line 2666: I think we should revert this too.  "whose" implies unnecessary anthropomorphism, which is worth avoiding.

```suggestion
          with a Stream Identifier field of 0x00, the recipient MUST respond with a <xref target="ConnectionErrorHandler">connection error</xref> of type PROTOCOL_ERROR.
```

martinthomson (approved):
Just three things (two to essentially revert).

Lukasa:
Suggestions applied

PR #1102, opened 2022-03-28
Run the tool I used to clean RFC 9000 on this one

martinthomson:
I had to remove the <bcp14> tags from the BCP 14 text, but the tool
restored them.

Other than that, there is no difference in output in the text files that
this generates, so I'm going to say that this was successful.

The source has a ton of whitespace changes, but that seems to be all.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -1,12 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
-
-<!DOCTYPE rfc [
- <!ENTITY nbsp    "&#160;">
- <!ENTITY zwsp   "&#8203;">
- <!ENTITY nbhy   "&#8209;">
- <!ENTITY wj     "&#8288;">
-]>
-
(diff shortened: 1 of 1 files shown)


draft-ietf-httpbis-http2bis.xml
```diff
@@ -769,7 +545,8 @@ HTTP Frame {
                   </g>
                 </g>
               </svg>
-            </artwork>
+<![CDATA[
+            ]]></artwork>
```
Lukasa, line 549: These `CDATA` insertions (here and above) are a bit of a surprise. Why are they here?

Lukasa (approved):
One note about the CDATA, this is otherwise fine.

draft-ietf-httpbis-http2bis.xml
martinthomson: Oh, that's a bug in the script.  Well, not so much a bug as a limitation that comes as a result of `<artwork>` sometimes containing ASCII art and other times containing `<svg>`.  Thanks for catching that, I'll fix it.

PR #1103, opened 2022-03-28
XML changes

martinthomson:
Mostly.

I didn't remove the width and height attributes from the SVG.  Removing
these causes the diagram to be stretched to fit, which is not at all
desirable (especially in the default stylesheet).

I restored Kazuho's name in the priority doc references and changed it
to use the allocated number of RFC 9218 for that doc.

I removed entities and used literals for the two &nbsp; instances in the
document (these are now &#xa0;).

I sorted the xi:include elements for references.

The file we got from the RFC editor included quoted values for URI
schemes and pseudo-header field names in the form: `<tt>":scheme"</tt>`.
This was not consistently applied.  I've gone with what we had:
`"<tt>:scheme</tt>"`.

This builds on top of the cleanup in #1102 as I have no interest in dealing with rebases over that change.

### DIFF
--- a/.note.xml
+++ b/.note.xml
@@ -1,7 +1,0 @@
-<note title="Discussion Venues" removeInRFC="true">
-<t>Discussion of this document takes place on the
-  HTTPBIS Working Group mailing list (ietf-http-wg@w3.org),
-  which is archived at <eref target="https://lists.w3.org/Archives/Public/ietf-http-wg/"/>.</t>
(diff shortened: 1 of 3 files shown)


Lukasa (approved):
One note about the CDATA, this is otherwise fine.

Lukasa (approved):

martinthomson:
I'll merge then fix to avoid problems.

PR #1104, opened 2022-05-30
Last minute editorial stuff

martinthomson:
The date will probably be June something, but let's worry about that
later.

### DIFF
--- a/draft-ietf-httpbis-http2bis.xml
+++ b/draft-ietf-httpbis-http2bis.xml
@@ -18,7 +18,7 @@
         <email>cbenfield@apple.com</email>
       </address>
     </author>
-    <date month="March" year="2022"/>
+    <date month="May" year="2022"/>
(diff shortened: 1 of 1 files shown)


Lukasa (approved):

draft-ietf-httpbis-http2bis.xml
```diff
@@ -3070,7 +3070,7 @@ cookie: e=f
         <reference anchor="RFC9112" target="https://www.rfc-editor.org/info/rfc9112">
           <front>
             <title>HTTP/1.1</title>
-            <author fullname="Roy T. Fielding" role="editor">
+            <author fullname="Roy Fielding" role="editor">
```
reschke, line 3073: The fullname accordng to RFC9112 is "Roy T. Fielding". The front matter disagrees. If you want to be ultra-correct, specify surname and initials separately.

